PR #63, opened 2021-03-16
Replacing commons-logging with slf4j in transitive dependencies

ricardozanini:
Signed-off-by: Ricardo Zanini <zanini@redhat.com>

Since the most used log framework is slf4j, and we are using it in our project, it doesn't make sense to bring another logging framework to our dependency list. One of our dependencies uses `commons-logging`. I just replaced it with a bridge to slf4j. So now, everything is logged using the same framework.

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -19,6 +19,10 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>jcl-over-slf4j</artifactId>
+        </dependency>
+        <dependency>
             <groupId>com.fasterxml.jackson.core</groupId>
             <artifactId>jackson-core</artifactId>
         </dependency>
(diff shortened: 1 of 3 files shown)


api/pom.xml
```diff
@@ -18,6 +18,10 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-api</artifactId>
         </dependency>
+        <dependency>
```
tsurdilo, line 21: Why is this depends needed? 
You could just exclude commons logging like you do below 
and have the client add whatever they might need

tsurdilo (changes requested):

api/pom.xml
ricardozanini: It's needed to give our transitive dependency the ability to log. Since they use `commons-logging`, it's likely that they require its interface. slf4j bridge does that: http://www.slf4j.org/legacy.html

tsurdilo: ok

tsurdilo (approved):

PR #89, opened 2021-05-04
(bugsbash) Remove dead code

soniasingla:


diagram/src/main/java/io/serverlessworkflow/diagram/utils/WorkflowDiagramUtils.java
```diff
@@ -32,7 +32,6 @@
     public static final String description = " : ";
     public static final String title = "title ";
     public static final String footer = "center footer Serverless Workflow Specification - serverlessworkflow.io";
-    public static final String legendStart = new StringBuffer().append("legend top center").append(System.lineSeparator()).toString();
```
tsurdilo: legendStart is used - see failing tests. i think this one should be kept and not removed

tsurdilo (changes requested):

diagram/src/main/java/io/serverlessworkflow/diagram/utils/WorkflowDiagramUtils.java
soniasingla: Updated! Thanks for the review :)

Issue #127, opened 2021-09-09
Add workflow utilities


tsurdilo:
**What would you like to be added**:
Add utility methods that implementors can use instead of having to write it themselves.
Ideas such as:
1. getStartingState
2. getStatesByType
3. getNextState
4. getConsumedEvents
5. getProducedEvents
6. getFunctionsDefsForAction
7. getActionsForFunctionDef
8. ...

manick02:
I will look at this one and come back

tsurdilo:
nice. thanks! idea is to create a top-level module like api/spi/validation etc and provide static util methods that take in a Workflow object as one of the params and return whatever is needed per util method.

manick02:
Do we have to create utils ? I feel the code is lot structured if added as part of API (module). We have a lot of test workflows and utility test methods already in API which we can re-use for our unit tests for the above requirement.

manick02:
@tsurdilo what does getNextState mean from runtime perspective? What is the functionality utility class should solve for here?

tsurdilo:
@manick02 good point. yeah i think no need to add that one.

manick02:
> @manick02 could we also add "getNumOfXYZ" methods, for example:
> 
> getNumOfStates() getNumOfStates(State.Type) getNumOfEndStates() getNumOfTransitions() getNumOfCompensationStates() getNumOfContinueAs() getNumOfActions() ... wdyt?
We can, Will runtime need these methods

tsurdilo:
i think we can close this now. we can open new issues for specific util methods that we want to add

PR #133, opened 2021-09-30
 Implemented Utility Method- GetConsumedEvents and GetProducedEvents #127

manick02:
Signed-off-by: manick02 <manickavasagam.sundaram@gmail.com>

Added utility method for get consumed and get produced events

### DIFF
--- a/utils/src/main/java/io/serverlessworkflow/utils/WorkflowUtils.java
+++ b/utils/src/main/java/io/serverlessworkflow/utils/WorkflowUtils.java
@@ -16,6 +16,7 @@
 package io.serverlessworkflow.utils;
 
 import io.serverlessworkflow.api.Workflow;
+import io.serverlessworkflow.api.events.EventDefinition;
 import io.serverlessworkflow.api.interfaces.State;
 import io.serverlessworkflow.api.start.Start;
 import io.serverlessworkflow.api.states.DefaultState;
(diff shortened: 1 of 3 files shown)


tsurdilo:
@manick02 thanks for the PR, one thing I wanted to mention:

There is a difference between events that are defined inside the event definitions array, and events that are actually 
consumed / produced inside your workflow control flow logic (states array).

So taken your PR, I would:
1) Define in the method names that its returning the *defined* events only, for example change "getConsumedEvents" to "getDefinedConsumedEvents", similar for the produced method name.
2) Also add count methods, for example "getDefinedConsumedEventsCount" (returning 0 if either no events defined or there are no consumed ones)

In addition I would add methods that return the actually produced and consumed events inside the workflow states array. 
For that you will need to iterate each workflow state and check:
1) For consumed, check each event definition, check each action that has eventRef, check each switch states event conditions, check each callback state, etc, for all those they should be consumed events (you can correlate back to the events array to get the type. Accumulate all those and return a list of unique ones (also add the size method for that please)
2) For produced events, check each state if it has end definition, and then check if it has produceEvents. You can check back into the events array to make sure that type is set to produced on those (this should really be part of the validation but whatever). Same here add a size method that returns the size of the list of those.

Does this make sense?

tsurdilo:
One last thing, for the methods that return the actual events used inside states array I would put a comment there 
that those only contain the events of the **parent** workflow (the workflow being evaluated), meaning that if there are any subflowRefs in the parent workflow being evaluated, the method DOES NOT go into the sub-workflow definitions to check those as well.

manick02:
Got it. It makes sense. I will update the PRs accordingly, could not anticipate what would be needed for runtime 


> Does this make sense?
> 
> You could break this up into 2 separate PRs if that makes it easier. If you chose that then the only change request here is to rename the methods as mentioned.

manick02:
> In addition I would add methods that return the actually produced and consumed events inside the workflow states array. (you could call them for example getWorkflowConsumedEvents" and "getWorkflowProducedEvents" for example)
> 
> For that you will need to iterate each workflow state and check:
> 
> 1. For consumed, check each event definition, check each action that has eventRef, check each switch states event conditions, check each callback state, etc, for all those they should be consumed events (you can correlate back to the events array to get the type. Accumulate all those and return a list of unique ones (also add the size method for that please)
this is lot :)
For Consumed Events : I need to consider below State types

- OnEvent 
- SwitchState and Callback State (as you specified)
Apart from the above, is there any other scenario i need to get evenRefs.  I was not sure how to get evenRef from action, as most of it I got it from the above States  @tsurdilo

tsurdilo:
@manick02
>  I was not sure how to get evenRef from action

get the action eventRef object and then get the triggerEventRef and resultEventRef. then you can look through the top-level events array and match by those values compare to event name

PR #196, opened 2022-04-22
Constanst.setRefValue() was always null

fjtirado:
Setting the ref value when constant is an uri to allow implementations
to try load them in case default loading fails

### DIFF
--- a/api/src/main/java/io/serverlessworkflow/api/serializers/WorkflowSerializer.java
+++ b/api/src/main/java/io/serverlessworkflow/api/serializers/WorkflowSerializer.java
@@ -160,10 +160,11 @@ public class WorkflowSerializer extends StdSerializer<Workflow> {
     }
 
     if (workflow.getConstants() != null) {
-      if (!workflow.getConstants().getConstantsDef().isEmpty()) {  
-          gen.writeObjectField("constants", workflow.getConstants().getConstantsDef());
+      if (workflow.getConstants().getConstantsDef() != null
+          && !workflow.getConstants().getConstantsDef().isEmpty()) {
+        gen.writeObjectField("constants", workflow.getConstants().getConstantsDef());
       } else if (workflow.getConstants().getRefValue() != null) {
-          gen.writeStringField("constants", workflow.getConstants().getRefValue());
+        gen.writeStringField("constants", workflow.getConstants().getRefValue());
       }
     }
 
--- a/api/src/main/java/io/serverlessworkflow/api/serializers/WorkflowSerializer.java
+++ b/api/src/main/java/io/serverlessworkflow/api/serializers/WorkflowSerializer.java
@@ -159,8 +159,12 @@ public class WorkflowSerializer extends StdSerializer<Workflow> {
       gen.writeEndArray();
     }
 
-    if (workflow.getConstants() != null && !workflow.getConstants().getConstantsDef().isEmpty()) {
-      gen.writeObjectField("constants", workflow.getConstants().getConstantsDef());
+    if (workflow.getConstants() != null) {
+      if (!workflow.getConstants().getConstantsDef().isEmpty()) {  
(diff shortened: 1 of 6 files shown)


ricardozanini (approved):
@tsurdilo I think we can set up a formatting profile for the SDK (see the imports). I'll do it. So, with a `mvn clean install` the files are automatically formatted.

fjtirado:
> @tsurdilo I think we can set up a formatting profile for the SDK (see the imports). I'll do it. So, with a `mvn clean install` the files are automatically formatted.

I remove the changed imports, but yes, template is a nice idea, thanks @ricardozanini

tsurdilo:
Thanks @fjtirado , could we please add a test for this change as well?

tsurdilo:
@manick02 can you please work with @fjtirado on this pr? Let me know when its good to merge. Would be nice to:
1. add a test for it
2. maybe we should have a null check for constantsFileDef
3. whatever else you see might be needed
Thanks!

manick02:
> @manick02 can you please work with @fjtirado on this pr? Let me know when its good to merge. Would be nice to:
> 
> 1. add a test for it
> 2. maybe we should have a null check for constantsFileDef
> 3. whatever else you see might be needed
>    Thanks!

sure @tsurdilo

manick02:
> @tsurdilo I think we can set up a formatting profile for the SDK (see the imports). I'll do it. So, with a `mvn clean install` the files are automatically formatted.

We have this plugin already in java sdk @ricardozanini 
           <plugin>
                    <groupId>com.coveo</groupId>
                    <artifactId>fmt-maven-plugin</artifactId>
                    <version>${version.fmt-maven-plugin}</version>
                </plugin>

fjtirado:
@tsurdilo 
Test added

api/src/test/java/io/serverlessworkflow/api/test/MarkupToWorkflowTest.java
```diff
@@ -569,6 +569,23 @@ public void testConstants(String workflowLocation) {
...
+    Workflow workflow = Workflow.fromSource(WorkflowTestUtils.readWorkflowFile(workflowLocation));
+
+    assertNotNull(workflow);
+    assertNotNull(workflow.getId());
+    assertNotNull(workflow.getName());
+    assertNotNull(workflow.getStates());
+
+    assertNotNull(workflow.getConstants());
+    Constants constants = workflow.getConstants();
+    assertNotNull(constants.getRefValue());
```
manick02: I think This assertion needs to be more strict to ensure the new behavior. recommend changing the assertion to assertEquals("  constantValues.json".constants.getRefValue());

api/src/main/java/io/serverlessworkflow/api/serializers/WorkflowSerializer.java
```diff
@@ -159,8 +159,12 @@ public void serialize(Workflow workflow, JsonGenerator gen, SerializerProvider p
       gen.writeEndArray();
     }
 
-    if (workflow.getConstants() != null && !workflow.getConstants().getConstantsDef().isEmpty()) {
-      gen.writeObjectField("constants", workflow.getConstants().getConstantsDef());
+    if (workflow.getConstants() != null) {
```
cb-manick, line 162: Is it possible to write test for this in WorkflowToMarkupTest?

fjtirado: sure, done

api/src/test/java/io/serverlessworkflow/api/test/MarkupToWorkflowTest.java
fjtirado: yes, better that way ;)

manick02:
@tsurdilo Looks Good from my side

tsurdilo:
@manick02 thanks!!

PR #199, opened 2022-05-09
New validations implemented

hbelmiro:
This is a work in progress...

# Validations Implemented

- [x] Switch state must contains dataCondition or eventCondition, not both
- [x] Switch state - Default condition - next state or end required
- [x] Switch state - Default condition - next state must be the name of an existing state
- [x] Switch state - Event conditions - timeout required
- [x] States - Event timeout - timeout must be a valid duration

# Final proposal

Validations are grouped by workflow definitions (state, even, function...).

# Initial proposal (not approved)

I'm implementing new workflow validations and wanted to check what do you guys think of my approach?
Currently, we’re performing all the validations in the `io.serverlessworkflow.validation.WorkflowValidatorImpl#validate` method. That method already is quite long, and my new validations would make it even longer. That would increase the code complexity and it would become harder to test.
So I decided to break the new validations into new classes.

I created this `io.serverlessworkflow.validation.WorkflowValidation` interface:
```java
interface WorkflowValidation {
  List<ValidationError> validate(Workflow workflow);
}
```

Each validation will implement that interface and be registered as a service on `resources/META-INF/services/io.serverlessworkflow.validation.WorkflowValidation`, so the original `WorkflowValidatorImpl#validate` method can load those services and perform the validations.

The current validations are still in the original method. I only implemented a couple of new validations in this new structure. If you like it, we can move all the validations to their respective classes in the future.

Advantages of this approach:

- The validation code is simpler and easier to read, since it only cares about what's being validated.
- The code is easier to test, since we don't need a full valid workflow to test a class. We can just set the workflow elements that matter for that test and we don't need to care about the other ones.
- Changes in one validation won't affect other validation tests.
    For instance: If we have a new required attribute in the spec, we just have to create a new validation and test it. We don't need to change all the other tests to include this new required attribute, because they don't care about this new attribute.

Disadvantage of this approach:
- We'll might lose a bit of performance, since we'll visit the same structure more than once. But, I don't believe this will be significant.

Please let me know what do you think and if I can continue implementing the validations this way. Feel free to suggest improvements in this solution as well.

### DIFF
--- a/validation/src/main/java/io/serverlessworkflow/validation/state/StatesValidator.java
+++ b/validation/src/main/java/io/serverlessworkflow/validation/state/StatesValidator.java
@@ -1,0 +1,64 @@
+/*
+ * Copyright 2022-Present The Serverless Workflow Specification Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.validation.state;
+
+import io.serverlessworkflow.api.Workflow;
+import io.serverlessworkflow.api.interfaces.State;
+import io.serverlessworkflow.api.states.DefaultState.Type;
+import io.serverlessworkflow.api.validation.ValidationError;
+import java.util.Collection;
+import java.util.EnumMap;
+import java.util.List;
+import java.util.Map;
+
+public final class StatesValidator {
+  private final Map<Type, StateValidator> stateValidators = new EnumMap<>(Type.class);
+
+  private final Collection<ValidationError> validationErrors;
+
+  public StatesValidator(Collection<ValidationError> validationErrors) {
+    this.validationErrors = validationErrors;
+
+    stateValidators.put(Type.EVENT, new EventStateValidator(validationErrors));
+    stateValidators.put(Type.OPERATION, new OperationStateValidator(validationErrors));
+    stateValidators.put(Type.SWITCH, new SwitchStateValidator(validationErrors));
+    stateValidators.put(Type.SLEEP, new SleepStateValidator(validationErrors));
+    stateValidators.put(Type.PARALLEL, new ParallelStateValidator(validationErrors));
+    stateValidators.put(Type.INJECT, new InjectStateValidator(validationErrors));
+    stateValidators.put(Type.FOREACH, new ForEachStateValidator(validationErrors));
+    stateValidators.put(Type.CALLBACK, new CallbackStateValidator(validationErrors));
+  }
+
+  public boolean hasValidStates(Workflow workflow) {
+    boolean finalValidationResult = true;
+
+    List<State> states = workflow.getStates();
+    if (states != null && !states.isEmpty()) {
+      for (State state : states) {
+        StateValidator stateValidator = stateValidators.get(state.getType());
+
+        if (stateValidator != null) {
+          finalValidationResult =
+              finalValidationResult && stateValidator.isValidState(workflow, state);
+        } else {
+          throw new IllegalStateException("No validator found for state type: " + state.getType());
+        }
+      }
+    }
+
+    return finalValidationResult;
+  }
+}
--- a/validation/README.md
+++ b/validation/README.md
@@ -1,0 +1,49 @@
+# Workflow Validations
+
+| Description                                 | Implemented |
+|---------------------------------------------|-------------|
+| `id` must be present and not be empty.      | Yes         |
+| `version` must be present and not be empty. | Yes         |
+| `start` must refer to an existing `state`.  | Yes         |
+| `states` must be present and not be empty.  | Yes         |
+
+
+## Functions
+
+| Description                                                    | Implemented |
+|----------------------------------------------------------------|-------------|
+| `operation` must be present.                                   | No          |
+
+## States
+
+| Description                              | Implemented |
+|------------------------------------------|-------------|
+| `name` must be present and not be empty. | Yes         |
+
+
+### Switch State
+
+| Description                                                    | Implemented |
+|----------------------------------------------------------------|-------------|
+| `dataCondition` or `eventCondition` must be present.           | Yes         |
+| Both `dataCondition` and `eventCondition` must not be present. | Yes         |
+| `defaultCondition` must not be present.                        | Yes         |
+
+#### Default Condition
+
+| Description                                    | Implemented |
+|------------------------------------------------|-------------|
+| `nextState` or `end` must be present.          | Yes         |
+| `nextState` must refer to an existing `state`. | Yes         |
+
+#### Event Condition
+
+| Description                                | Implemented |
+|--------------------------------------------|-------------|
+| Event or workflow timeout must be present. | Yes         |
+
+#### Timeouts
+
+| Description                                       | Implemented |
+|---------------------------------------------------|-------------|
+| `eventTimeout` must be a valid ISO 8601 duration. | Yes         |
--- a/validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
+++ b/validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
@@ -18,17 +18,23 @@ package io.serverlessworkflow.validation;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.fasterxml.jackson.dataformat.yaml.YAMLFactory;
 import io.serverlessworkflow.api.Workflow;
-import io.serverlessworkflow.api.actions.Action;
 import io.serverlessworkflow.api.events.EventDefinition;
 import io.serverlessworkflow.api.events.OnEvents;
 import io.serverlessworkflow.api.functions.FunctionDefinition;
 import io.serverlessworkflow.api.interfaces.State;
 import io.serverlessworkflow.api.interfaces.WorkflowValidator;
-import io.serverlessworkflow.api.states.*;
+import io.serverlessworkflow.api.states.CallbackState;
+import io.serverlessworkflow.api.states.EventState;
+import io.serverlessworkflow.api.states.ForEachState;
+import io.serverlessworkflow.api.states.InjectState;
+import io.serverlessworkflow.api.states.ParallelState;
+import io.serverlessworkflow.api.states.SleepState;
+import io.serverlessworkflow.api.states.SwitchState;
 import io.serverlessworkflow.api.switchconditions.DataCondition;
 import io.serverlessworkflow.api.switchconditions.EventCondition;
 import io.serverlessworkflow.api.validation.ValidationError;
 import io.serverlessworkflow.api.validation.WorkflowSchemaLoader;
+import io.serverlessworkflow.validation.state.StatesValidator;
 import java.util.ArrayList;
 import java.util.HashSet;
 import java.util.List;
@@ -43,7 +49,10 @@ public class WorkflowValidatorImpl implements WorkflowValidator {
 
   private static final Logger logger = LoggerFactory.getLogger(WorkflowValidatorImpl.class);
   private boolean schemaValidationEnabled = true;
-  private List<ValidationError> validationErrors = new ArrayList<>();
+  private final List<ValidationError> validationErrors = new ArrayList<>();
+
+  private final StatesValidator statesValidator = new StatesValidator(validationErrors);
+
   private Schema workflowSchema = WorkflowSchemaLoader.getWorkflowSchema();
   private String source;
   private Workflow workflow;
@@ -121,6 +130,8 @@ public class WorkflowValidatorImpl implements WorkflowValidator {
             "Workflow version should not be empty", ValidationError.WORKFLOW_VALIDATION);
       }
 
+      statesValidator.hasValidStates(workflow);
+
       if (workflow.getStates() == null || workflow.getStates().isEmpty()) {
         addValidationError("No states found", ValidationError.WORKFLOW_VALIDATION);
       }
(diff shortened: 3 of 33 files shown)


tsurdilo:
Does this division play nicely with service loader? https://github.com/serverlessworkflow/sdk-java/blob/main/spi/src/test/resources/META-INF/services/io.serverlessworkflow.api.interfaces.WorkflowValidator

I am honestly not a big fan of splitting up each validation into own class, think long term it will just create more maintenance. 
If we have to split this up would prefer a more "natural" division based on workflow def, for example "states", "timeouts", "function def", "events def" ... but still keep a top-level validation impl that can group these together. 
@manick02 wdyt?

Also, if you need extra validation ontop of what SDK provides another option is to define your own validator and set it via service loader.

manick02:
I agree with @tsurdilo 
We can have one interface at the top level (Workflow) and the top level implementation of it decides how to validate sub parts of workflow(states, timeouts) etc. The proposed level of subpart validations are very granular to my liking.

hbelmiro:
This is ready to review, @tsurdilo.

validation/src/main/java/io/serverlessworkflow/validation/StatesValidator.java
```diff
@@ -0,0 +1,209 @@
...
+  private StatesValidator() {}
+
+  static List<ValidationError> validateStates(Workflow workflow) {
+    List<State> states = workflow.getStates();
+
+    if (states != null && !states.isEmpty()) {
+      List<ValidationError> errors = new ArrayList<>();
+      for (State state : states) {
+        if (state instanceof SwitchState) {
+          errors.addAll(validateSwitchState(workflow, (SwitchState) state));
```
fjtirado: @hbelmiro I think a more OO approach (avoiding instanceof) will be to define a Validator interface and a SwichValidator class that implement it
So the code will look like
`for (State state: states) { validators.get(state.getType()).validate(workflow,state)}`

fjtirado: and then the code will be spread into different state validator rather than having a huge StatesValidator

tsurdilo:
@hbelmiro thanks for the updates. will review as soon as possible.

ricardozanini (commented):
So are we proceeding with the validations of the other sections such as functions, events, and so on in other PRs?

tsurdilo:
@hbelmiro so this is just validating switch state at this time?

hbelmiro:
@tsurdilo @ricardozanini 
Actually I was thinking of implement on demand. But since you prefer to merge everything together, I think we can do in one single PR.
I'll revert it to draft and continue with the implementation.
Thanks for now, guys.

validation/src/main/java/io/serverlessworkflow/validation/StatesValidator.java
hbelmiro: Good idea. Thanks @fjtirado.

hbelmiro:
> Just really wanna make sure it plays well with service loader in the end.

@tsurdilo currently only the `WorkflowValidatorImpl` class is a service. The section validations are in a static method which is called by `WorkflowValidatorImpl`.
We can turn the section validations into services (rather than use static methods) if you prefer. If we decide to do that, we have two alternatives:

1. Create specific interfaces for each section. This gives the user flexibility to specify implementations for each section.

2. All section validations implementing `WorkflowValidatorImpl`. Simpler, but not so flexible. Also, the section implementations would have to implement `WorkflowValidatorImpl`'s methods that don't make much sense for section implementations, like:

```java
WorkflowValidator setWorkflow(Workflow workflow);
WorkflowValidator setSchemaValidationEnabled(boolean schemaValidationEnabled);
```

tsurdilo:
@hbelmiro thanks for the info! what do you think would be better? 
my best guess right now and could be wrong is to just leave the top-level WorkflowValidatorImpl and allow users to set their own if they want via config. if there is a strong need for being able to overwrite only certain groups of validations then your approach would be better imo, so I'm torn :) 

In my mind having users extend our WorkflowValidatorImpl and be able to override a particular validation method would work too. wdyt?

hbelmiro:
@tsurdilo I think we can go on the way we are (users can set their own WorkflowValidator only). In the future we can make it more flexible if needed.

validation/src/main/java/io/serverlessworkflow/validation/StatesValidator.java
hbelmiro: I would still need to cast `state` to `SwitchState` to have access to its methods.

validation/src/main/java/io/serverlessworkflow/validation/CommonStateValidator.java
```diff
@@ -0,0 +1,41 @@
...
+        T typedState = (T) state;
+        List<ValidationError> errors = runSpecificValidations(workflow, typedState);
+
+        validateStateName(state).ifPresent(errors::add);
+
+        return errors;
+    }
+
+    private Optional<ValidationError> validateStateName(State state) {
+        if (state.getName() != null && state.getName().trim().isEmpty()) {
```
fjtirado: I think there is a typo here, it should be !state.getName().trim().isEmpty()

validation/src/main/java/io/serverlessworkflow/validation/CommonStateValidator.java
```diff
@@ -0,0 +1,41 @@
...
+    }
+
+    protected final ValidationError error(String message) {
+        ValidationError error = new ValidationError();
+        error.setType(ValidationError.WORKFLOW_VALIDATION);
+        error.setMessage(message);
+        return error;
+    }
+
+    protected abstract List<ValidationError> runSpecificValidations(Workflow workflow, T state);
```
fjtirado: Rather than declaring this method as abstract (you can still declare the class as abstract event if you do not have any abstrace method), you can return new ArrayList by default and save a lot of code in childred classes that are just returning new ArrayList

validation/src/main/java/io/serverlessworkflow/validation/StateValidator.java
```diff
@@ -0,0 +1,12 @@
...
+
+import io.serverlessworkflow.api.Workflow;
+import io.serverlessworkflow.api.interfaces.State;
+import io.serverlessworkflow.api.validation.ValidationError;
+
+import java.util.List;
+
+interface StateValidator {
+
+    List<ValidationError> validate(Workflow workflow, State state);
```
fjtirado: Unless there is need to random access, I will declare this as collection rather than list (you might eventually implement equals in ValidationError and use a set without changing the interface)

validation/src/main/java/io/serverlessworkflow/validation/StatesValidator.java
```diff
@@ -16,194 +16,44 @@
...
-
-  static List<ValidationError> validateStates(Workflow workflow) {
-    List<State> states = workflow.getStates();
-
-    if (states != null && !states.isEmpty()) {
-      List<ValidationError> errors = new ArrayList<>();
-      for (State state : states) {
-        if (state instanceof SwitchState) {
-          errors.addAll(validateSwitchState(workflow, (SwitchState) state));
+    private static final EnumMap<Type, StateValidator> STATE_VALIDATORS = new EnumMap<>(Map.of(
```
fjtirado: This is nit pick, but I think the Java naming convention states that final static references should not be in UPPERCASE.

fjtirado: Althoug you are using EnumMap as implementation, you should use the interface as variable type, Map in this case

fjtirado: Rather than using the copy constructor, it will be better to use `new EnumMap(Type.class)` and use a static block to initialize it (calling several put)  rather than Map.of.

validation/src/main/java/io/serverlessworkflow/validation/StatesValidator.java
```diff
@@ -16,194 +16,44 @@
...
+
+        if (states != null && !states.isEmpty()) {
+            List<ValidationError> errors = new ArrayList<>();
+            for (State state : states) {
+                StateValidator stateValidator = STATE_VALIDATORS.get(state.getType());
+
+                if (stateValidator != null) {
+                    errors.addAll(stateValidator.validate(workflow, state));
+                } else {
+                    throw new IllegalStateException("No validator found for state type: " + state.getType());
```
fjtirado: I think you are not providing the complete set of validator. hence this will eventually throw IlleglStateException even for valid states.  
I would rather do nothing if there is not validator for that state and print a log.

fjtirado: Also by doing that, you wont need dummy implementations of validator that just return an empty list of errors (they can be added lated when the validator is implemented)

validation/src/main/java/io/serverlessworkflow/validation/SwitchStateValidator.java
```diff
@@ -0,0 +1,163 @@
...
+
+        return errors;
+    }
+
+    private static List<ValidationError> validateDataConditionAndEventConditionPresent(
+            SwitchState switchState) {
+        List<ValidationError> errors = new ArrayList<>();
+
+        if (!switchState.getDataConditions().isEmpty() && !switchState.getEventConditions().isEmpty()) {
+            ValidationError error = new ValidationError();
```
fjtirado: You already have a protected final method in the parent class to create the validation error

validation/src/main/java/io/serverlessworkflow/validation/SwitchStateValidator.java
```diff
@@ -0,0 +1,163 @@
...
+        return Optional.ofNullable(workflow.getTimeouts()).map(TimeoutsDefinition::getEventTimeout);
+    }
+
+    private Optional<ValidationError> validateNextState(
+            Workflow workflow, String switchStateName, DefaultConditionDefinition defaultCondition) {
+        String nextState =
+                defaultCondition.getTransition() != null
+                        ? defaultCondition.getTransition().getNextState()
+                        : null;
+        if (nextState != null && !nextState.isEmpty()) {
```
fjtirado: I think it will be nice to have an utility static method that check if a string is not null and is not empty, so you save a lot of code and dummy erros (in this case you forget the trim, in other case you miss the !)

fjtirado (commented):
Some things that need to be sorted out specifically commented in the code. 
Although I do not mention that specifically, in general, I think we are using Optional too much. 
I understood the reason, to not pass the list of error to all the validation methods, and do not have this kind of idiom
`if (validateXXX) {errors.add(xxx_ERROR)}`
Having
`validateXXX().ifPresent(errors:add);`
you have the error condidition and the error message in the validateXXX(), but since at the end the error message is a constant, I found easier to read the first one.

validation/src/main/java/io/serverlessworkflow/validation/StatesValidator.java
fjtirado: so it should be named `stateValidators` (easier to read lated when the map is used)

hbelmiro:
Thanks for the review @fjtirado.
I'm still working on it. I just pushed yesterday as a backup. So, there are several parts that won't stay as they are and some are not even working. Sorry for wasting your time.
However, some reviews are already useful, like the typo, the EnumMap one, and others. So, it wasn't a total waste of time :)

Regarding the use of Optional, I can remove it. In general, I like using Optional (always trying to not overuse it) to let things obvious and more intuitive.

tsurdilo:
@hbelmiro 

Screen Shot 2022-05-27 at 5 16 27 PM

could we please not mention KOGITO (your product) in commit names?
this sdk or this project has anything to do w/ it :)

I would actually like if possible to request that this pr be redone w/ commits not mentioning that project, but specifically what they change within this SDK and what changes to this sdk they include. Thank you!

hbelmiro:
> could we please not mention KOGITO in commit names?
> this sdk is not at all related to this project.
> 
> I would like to request to update the commits if possible to just include what the changes are. Thanks!!

Sure. I'll fix them.

validation/src/main/java/io/serverlessworkflow/validation/state/StatesValidator.java
```diff
@@ -0,0 +1,64 @@
...
+    boolean finalValidationResult = true;
+
+    List<State> states = workflow.getStates();
+    if (states != null && !states.isEmpty()) {
+      for (State state : states) {
+        StateValidator stateValidator = stateValidators.get(state.getType());
+
+        if (stateValidator != null) {
+          finalValidationResult =
+              finalValidationResult && stateValidator.isValidState(workflow, state);
```
fjtirado, line 55: so, at the moment a validator fail, we cancel the whole validation process?
I guess this is intentional, but just double checking. 
Cancelling the validation process when the first state that has a validation error is found will make the validationErrors list not complete.

validation/src/main/java/io/serverlessworkflow/validation/state/StatesValidator.java
```diff
@@ -0,0 +1,64 @@
...
+
+public final class StatesValidator {
+  private final Map<Type, StateValidator> stateValidators = new EnumMap<>(Type.class);
+
+  private final Collection<ValidationError> validationErrors;
+
+  public StatesValidator(Collection<ValidationError> validationErrors) {
+    this.validationErrors = validationErrors;
+
+    stateValidators.put(Type.EVENT, new EventStateValidator(validationErrors));
```
manick02, line 35: Do we have to create the instance of validator for event type every we want to validate states?

validation/src/main/java/io/serverlessworkflow/validation/state/StatesValidator.java
hbelmiro: No. The idea is to return false, which means "doesn't have valid states", but continue with the validation.
Currently the return value is being ignored, but that's just because it's not fully implemented yet.

validation/src/main/java/io/serverlessworkflow/validation/state/StatesValidator.java
hbelmiro: Actually, it will create only one instance of `StatesValidator` per `WorkflowValidatorImpl` instance.
See: https://github.com/hbelmiro/sdk-java/blob/cdf134a2d324310e7fa37f6e63823019a104d4d4/validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java#L54
That instance of `StatesValidator` has several instances of `StateValidator` (one per type) - I know I need to think about better names :)
In the end, only one instance of each class will be created.

hbelmiro: Actually, it will create only one instance of `StatesValidator` per `WorkflowValidatorImpl` instance.
See: https://github.com/hbelmiro/sdk-java/blob/cdf134a2d324310e7fa37f6e63823019a104d4d4/validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java#L54
That instance of `StatesValidator` has several instances of `StateValidator` (one per type) - I know I need to think about better names :)
In the end, only one instance of each class will be created.

validation/src/main/java/io/serverlessworkflow/validation/state/StatesValidator.java
fjtirado: Ok, then, you need to change this code because stateValidator.isValidState(workflow, state) is not going to be invoked if previous state returned false. 
you should write` finalValidationResult&=stateValidator.isValidState(workflow, state);`

hbelmiro:
Before proceeding with this, I opened [a new PR to use Bean Validations](https://github.com/serverlessworkflow/sdk-java/pull/208). With that PR we'll significantly reduce the amount of manually implemented validations.

hbelmiro:
@ricardozanini closing as stale.

PR #203, opened 2022-06-18
Added new property to datainputschema.json

VaniHaripriya:
Added new property "schemaContent" to dataInputSchema.json to fix the issue with schema returning null when inline.

### DIFF
--- a/api/src/main/resources/schema/datainputschema/datainputschema.json
+++ b/api/src/main/resources/schema/datainputschema/datainputschema.json
@@ -8,6 +8,11 @@
       "description":  "URI of the JSON Schema used to validate the workflow data input",
       "minLength": 1
     },
+    "schemaContent":  {
+      "type": "object",
+      "description": "The JSON Schema object used to validate the workflow data input",
+      "existingJavaType": "com.fasterxml.jackson.databind.JsonNode"
+    },
     "failOnValidationErrors": {
       "type": "boolean",
       "default": true,
@@ -15,7 +20,6 @@
     }
   },
   "required": [
-    "schema",
     "failOnValidationErrors"
   ]
 }


tsurdilo:
@VaniHaripriya thanks for the pr
unfortunately I am not able to accept it:
1. not sure how it is related to specification schemas
2. no associated issue
3. no tests or explanation why this is needed

could you please open an issue first to describe what your change is about and then we can look at reviving this pr if needed

Thanks!

fjtirado:
@tsurdilo, this is a bug in the sdk, which does not obey the spec for DataInputSchema, currently the SDK only supports URI, while the spec support both URI and inlining. 
Indeed, there was a controversial issue and we agree to change the SDK because we want to support inlining (I was advocating for removing inlining support) 
https://github.com/serverlessworkflow/specification/issues/624

```
The dataInputSchema property can be used to validate the workflow data input against a defined JSON Schema. This check should be done before any states are executed. dataInputSchema can have two different types. If it is an object type it has the following definition:

"dataInputSchema": {
   "schema": "URL_to_json_schema",
   "failOnValidationErrors": false
}

It's schema property can be an URI, which points to the JSON schema used to validate the workflow data input, or it can be the JSON schema object. If it's a JSON schema object, it has the following definition:

"dataInputSchema": {
   "schema": {
     "title": "MyJSONSchema",
     "properties":{
       "firstName":{
         "type": "string"
       },
       "lastName":{
         "type": "string"
       }
     }
   },
   "failOnValidationErrors": false
}

```

I agree the PR cannot be accepted, though
@VaniHaripriya 
You need to add code to handle the new fields in the deserializer
https://github.com/serverlessworkflow/sdk-java/blob/main/api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java#L53

And also a test. 

I suggest reopening the PR and completing the missing bits.

ricardozanini:
@VaniHaripriya I agree with their comments, ideally, the SDK must conform with the spec, that's the missing piece.

fjtirado:
@ricardozanini 
This should be handled like the [Constants](https://github.com/serverlessworkflow/sdk-java/blob/main/api/src/main/java/io/serverlessworkflow/api/workflow/Constants.java) object

cdavernas:
@fjtirado I dont see why you think we need to add a `SchemaContent` property. The actual spec works perfectly without that property, even with the inline schema.

```
{
  "id": "order-dog",
  "version": "1.0.0",
  "specVersion": "0.8",
  "name": "Order pet",
  "description": "A workflow used to order a pet on the Swagger pet store",
  "dataInputSchema":{
      "schema":{
          "type":"object",
          "properties":{
              "username":{
                  "type": "string",
                  "default": "string"
              },
              "quantityToOrder":{
                  "type": "number",
                  "default": 3
              }
          }
      }
    },
    ...
}
```

fjtirado:
@cdavernas  It does not work with JAVA, schema field of JAVA object is null when inlining
As I sadi, I do NOT want to add any schemContent property to the spec, I want to add it to the generated JAVA object.

cdavernas:
@fjtirado Fair enough. It however feels like a dirty trick to add that property. Can't something be done to avoid that null ref alltogether and make a proper use of the `schema` field, even if that implies breaking backward compatibility (which anyways should never be assumed before an RC, IMO)?

fjtirado:
@cdavernas The reason I suggest the additional field in the Java object is because is aligned with what has been done in other cases I have referenced in my previous comments, for example Constants.

cdavernas:
@fjtirado Fine by me then! Had a look at it just now, btw, and realized it's a private, hidden, backing field. I therefore don't see any harm in adding a similar field in the Java object.

tsurdilo:
i think we can do same thing we did for other inline props. no problem

tsurdilo:
i would it rather be converted to jsonobject if possible iirc there shoild be code already we can reuse

ricardozanini:
@tsurdilo agreed

PR #205, opened 2022-06-22
Created a new POJO for DataInputSchema to support inline schema

VaniHaripriya:
Currently SDK is not supporting the inline schema as mentioned in the spec. It is supporting only URI.
Created a new pojo for DataInputSchema to handle the same. Also, made the related changes to DataInputSchemaDeserializer and created supporting tests.

### DIFF
--- a/api/src/main/java/io/serverlessworkflow/api/mapper/BaseObjectMapper.java
+++ b/api/src/main/java/io/serverlessworkflow/api/mapper/BaseObjectMapper.java
@@ -15,10 +15,12 @@
  */
 package io.serverlessworkflow.api.mapper;
 
+import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.core.JsonFactory;
 import com.fasterxml.jackson.databind.ObjectMapper;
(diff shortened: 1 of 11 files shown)


api/src/main/java/io/serverlessworkflow/api/mapper/BaseObjectMapper.java
```diff
@@ -31,6 +33,11 @@ public BaseObjectMapper(JsonFactory factory, WorkflowPropertySource workflowProp
 
     configure(SerializationFeature.INDENT_OUTPUT, true);
     registerModule(workflowModule);
+    configure(SerializationFeature.WRITE_EMPTY_JSON_ARRAYS, false);
```
manick02, line 36: this change is relevant to the fix?  seems to be a rebase from another bug fix that @tsurdilo  fixed recently

VaniHaripriya: @manick02  Yes, this was not related to the fix. I will create a new PR to avoid confusion.

VaniHaripriya:
Creating a new one, so closing this PR.

Issue #204, opened 2022-06-22
POJOs being generated with initialized collections by default


hbelmiro:
POJOs are being generated with initialized collections by default. This causes properties not set in the workflow to be unmarshaled to empty collections.

For instance, given the following JSON:

```json
{
  "id": "helloworld",
  "annotations": ["hello", "world"]
}
```

`getAnnotations()` should return a collection with two objects. This is working as expected.

Given the followin JSON:

```json
{
  "id": "helloworld",
  "annotations": []
}
```

`getAnnotations()` should return an empty collection. This is also working as expected.

Given the followin JSON: 

```json
{
  "id": "helloworld"
}
```

`getAnnotations()` should return `null`, but currently it's returning an empty collection. Here's the problem.

This is a problem because there are properties that aren't required, but when present, they should not be empty.
Translating this to Java:
| Value | Valid |
| --------| ------- |
| `null` | Yes |
| empty | No |
| not empty | Yes |

To fix this behavior, we just need to set [initializeCollections](https://joelittlejohn.github.io/jsonschema2pojo/site/1.1.2/generate-mojo.html#initializeCollections) to `false` in [pom.xml](https://github.com/serverlessworkflow/sdk-java/blob/8a61e3cb23f6ae75d2d711410bff974322717d77/api/pom.xml#L94).

**But, this change can break compatibility, since methods that today never return `null` might return `null` after the change, which may cause `NullPointerExceptions` to users.**

So, what do you guys think? Can I submit a PR to fix this issue?

tsurdilo:
i think this was fixed via https://github.com/serverlessworkflow/sdk-java/commit/8a61e3cb23f6ae75d2d711410bff974322717d77

hbelmiro:
It wasn't.
We can confirm that in [this assertion](https://github.com/serverlessworkflow/sdk-java/blob/8a61e3cb23f6ae75d2d711410bff974322717d77/api/src/test/java/io/serverlessworkflow/api/test/MarkupToWorkflowTest.java#L232). It should be failing, since `produceEvents` is not defined in the workflow.

```java
assertNotNull(defaultDefinition.getTransition().getProduceEvents());
```

```json
      "defaultCondition": {
        "transition": {
          "nextState": "RejectApplication",
          "compensate": true
        }
      }
```

tsurdilo:
| getAnnotations() should return null, but currently it's returning an empty collection

[...]

PR #206, opened 2022-06-23
Change to not initialize collections by default on generated POJOs

hbelmiro:
Fixes https://github.com/serverlessworkflow/sdk-java/issues/204.

Changes:
* Set [initializeCollections](https://joelittlejohn.github.io/jsonschema2pojo/site/1.1.2/generate-mojo.html#initializeCollections) jsonschema2pojo property to `false`
* Added test
* Fixed test that started to fail after setting `initializeCollections` to `false`

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -97,6 +97,7 @@
                     <outputDirectory>${project.build.directory}/generated-sources/src/main/java</outputDirectory>
                     <includeJsr303Annotations>true</includeJsr303Annotations>
                     <generateBuilders>true</generateBuilders>
+                    <initializeCollections>false</initializeCollections>
                     <includeAdditionalProperties>false</includeAdditionalProperties>
                     <includeToString>false</includeToString>
                     <includeHashcodeAndEquals>false</includeHashcodeAndEquals>
--- a/api/src/test/java/io/serverlessworkflow/api/test/CodegenTest.java
+++ b/api/src/test/java/io/serverlessworkflow/api/test/CodegenTest.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2020-Present The Serverless Workflow Specification Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.api.test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+import io.serverlessworkflow.api.Workflow;
+import io.serverlessworkflow.api.test.utils.WorkflowTestUtils;
+import org.junit.jupiter.api.Test;
+
+class CodegenTest {
+
+  @Test
+  void collectionsShouldNotBeInitializedByDefault() {
+    Workflow workflow =
+        Workflow.fromSource(WorkflowTestUtils.readWorkflowFile("/features/functionrefs.json"));
+    assertThat(workflow.getAnnotations()).isNull();
+  }
+}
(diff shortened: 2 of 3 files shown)


tsurdilo:
Can you show what the resulting json (Workflow.toJson(workflow)) looks like with this change?

hbelmiro:
When `initializeCollections = true` (current code):

```json
{
  "id" : "functionrefs",
  "name" : "Customer Credit Check Workflow",
  "description" : "Perform Customer Credit Check",
  "version" : "1.0",
  "start" : "TestFunctionRef",
  "specVersion" : "0.8",
  "expressionLang" : "jq",
  "events" : [ ],
  "functions" : [ {
    "name" : "creditCheckFunction",
    "operation" : "http://myapis.org/creditcheckapi.json#doCreditCheck",
    "type" : "rest"
  }, {
    "name" : "sendRejectionEmailFunction",
    "operation" : "http://myapis.org/creditcheckapi.json#rejectionEmail",
    "type" : "rest"
  } ],
  "retries" : [ ],
  "errors" : [ ],
  "secrets" : [ ],
  "states" : [ {
    "actionMode" : "sequential",
    "actions" : [ {
      "functionRef" : "creditCheckFunction"
    }, {
      "functionRef" : {
        "refName" : "sendRejectionEmailFunction",
        "arguments" : {
          "applicant" : "${ .customer }"
        },
        "invoke" : "sync"
      }
    } ],
    "usedForCompensation" : false,
    "name" : "TestFunctionRefs",
    "type" : "operation",
    "end" : true
  } ]
}
```

When `initializeCollections = false` (this change):

```json
{
  "id" : "functionrefs",
  "name" : "Customer Credit Check Workflow",
  "description" : "Perform Customer Credit Check",
  "version" : "1.0",
  "start" : "TestFunctionRef",
  "specVersion" : "0.8",
  "events" : [ ],
  "functions" : [ {
    "name" : "creditCheckFunction",
    "operation" : "http://myapis.org/creditcheckapi.json#doCreditCheck",
    "type" : "rest"
  }, {
    "name" : "sendRejectionEmailFunction",
    "operation" : "http://myapis.org/creditcheckapi.json#rejectionEmail",
    "type" : "rest"
  } ],
  "retries" : [ ],
  "errors" : [ ],
  "secrets" : [ ],
  "states" : [ {
    "actionMode" : "sequential",
    "actions" : [ {
      "functionRef" : "creditCheckFunction"
    }, {
      "functionRef" : {
        "refName" : "sendRejectionEmailFunction",
        "arguments" : {
          "applicant" : "${ .customer }"
        },
        "invoke" : "sync"
      }
    } ],
    "usedForCompensation" : false,
    "name" : "TestFunctionRefs",
    "type" : "operation",
    "end" : true
  } ]
}
```

There's only one difference: `"expressionLang" : "jq"` didn't appear in the json when `initializeCollections = false`
This shouldn't happen. If we'll go on with this change, I'll have to investigate it. But this is a different issue.
The most important thing, for what we're discussing, is that `annotations` doesn't appear in json no matter the value of `initializeCollection`. This behavior is correct when `initializeCollections = false` (this change), since `Workflow#getAnnotations` returns `null`. But when `initializeCollections = true`, `Workflow#getAnnotations` returns an empty list. So an empty array should appear in the resulting json. It looks like the json parser ignores the attribute if it's `null` or empty.

tsurdilo:
https://github.com/serverlessworkflow/sdk-java/blob/8a61e3cb23f6ae75d2d711410bff974322717d77/api/src/main/java/io/serverlessworkflow/api/serializers/WorkflowSerializer.java#L66

tsurdilo:
if its null or empty it should not be spit out as json/yaml

tsurdilo:
we need tests that round-trip markup it seems ...will add

hbelmiro:
Reverted to draft to investigate the jq issue.

tsurdilo:
> There's only one difference: "expressionLang" : "jq" didn't appear in the json

jq is the default value if its not in json/yaml so this should be ok imo

hbelmiro:
> > There's only one difference: "expressionLang" : "jq" didn't appear in the json
> 
> jq is the default value if its not in json/yaml so this should be ok imo
> 
> the markup that sdk spits out from object model imo does not 100% have to match the original user markup (for example default prop values and empty arrays can be omitted) , but it has to be semantically the same (and it has to be validatable if original was too)

Yes. Makes sense.

hbelmiro:
> we need tests that round-trip markup it seems ...will add

@tsurdilo Just to clarify, you need to add these tests before we move on with this PR?

tsurdilo:
> > we need tests that round-trip markup it seems ...will add
> 
> @tsurdilo Just to clarify, you need to add these tests before we move on with this PR?

Yes, I'm updating all tests to make sure the json+yaml outputs are validated. This could help us figure out all possible issues the code might have right now and then we can look how to handle them. Hope thats ok with you.

hbelmiro:
@tsurdilo did you manage to update the tests you mention above so we can move this PR forward?
If you need any help with that, just let me know.

hbelmiro:
@manick02 do you mind taking a look at this PR? I'm not sure if @tsurdilo is seeing my messages.
If you have any questions or need any clarification on this, please don't hesitate to ask me.

tsurdilo (approved):

PR #207, opened 2022-06-24
extension identifier should be extensionId

fjtirado:
The deserializer has a typo, fixing it


fjtirado:
@tsurdilo 
This will be difficult to test without adding a dummy extension in test. Do we wanto to add it? I guess the answer is yes ;), but just double checking.

tsurdilo:
Ah, i see. Ok lets merge this then i guess and i can add proper test for extensions.

PR #209, opened 2022-06-29
Created a new POJO for DataInputSchema and the required changes to support inline schema

VaniHaripriya:
This PR is to make changes to support inline schema for DataInputSchema.
Currently SDK is not supporting the inline schema as mentioned in the spec. It is supporting only URI.
Created a new pojo for DataInputSchema to handle the same. Also, made the corresponding changes related to this and created supporting tests.

Below is the related issue :
https://github.com/serverlessworkflow/specification/issues/624


api/src/main/resources/schema/workflow.json
```diff
@@ -36,7 +36,8 @@
       }
     },
     "dataInputSchema": {
-      "$ref": "datainputschema/datainputschema.json",
+      "type": "object",
```
manick02, line 39: @tsurdilo can you check this. I cannot think of an alternative but Is this approach fine? this would lead us to maintain java SDK specific workflow schema

ricardozanini: @manick02 if you see the other types (states, functions, secrets, for example), we are already doing that.

ricardozanini:
@tsurdilo mind taking a look, please?

api/src/main/resources/schema/workflow.json
tsurdilo: i think this is same thing we did with functions as well so maybe its ok

tsurdilo:
on first look looks good to me. id like to add some more tests locally and test it out before merging . ty for pr!

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,43 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
+      String schemaFileSrc = Utils.getResourceFileAsString(schemaFileDef);
+      JsonNode schemaRefNode;
+      ObjectMapper jsonWriter = new ObjectMapper();
+      if (schemaFileSrc != null && schemaFileSrc.trim().length() > 0) {
+        // if its a yaml def convert to json first
+        if (!schemaFileSrc.trim().startsWith("{")) {
+          // convert yaml to json to validate
+          ObjectMapper yamlReader = new ObjectMapper(new YAMLFactory());
+          Object obj = yamlReader.readValue(schemaFileSrc, Object.class);
+
```
manick02: nit:blank line

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,43 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
-    } else {
-      dataInputSchema.setSchema(node.get("schema").asText());
+    if (node.isObject() && node.get("schema").isObject() && !node.get("schema").isEmpty()) {
+      schemaDefinition = node.get("schema");
       dataInputSchema.setFailOnValidationErrors(node.get("failOnValidationErrors").asBoolean());
+    } else {
+      String schemaFileDef = node.isObject() ? node.get("schema").asText() : node.asText();
+      dataInputSchema.setFailOnValidationErrors(true);
+      dataInputSchema.setRefValue(schemaFileDef);
+      String schemaFileSrc = Utils.getResourceFileAsString(schemaFileDef);
```
manick02: Can we have a test for schemaFileDef if not physically present?  I cannot determine what should be the behavior we would want the code to exihibit

manick02: nit : renaming schemaFileSrc to schemaFileContent helps in understanding the code flow

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,43 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
+      JsonNode schemaRefNode;
+      ObjectMapper jsonWriter = new ObjectMapper();
+      if (schemaFileSrc != null && schemaFileSrc.trim().length() > 0) {
+        // if its a yaml def convert to json first
+        if (!schemaFileSrc.trim().startsWith("{")) {
+          // convert yaml to json to validate
+          ObjectMapper yamlReader = new ObjectMapper(new YAMLFactory());
+          Object obj = yamlReader.readValue(schemaFileSrc, Object.class);
```
manick02: nit: obj to more meaningful name

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,43 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
+
+          schemaRefNode =
+              jsonWriter.readTree(new JSONObject(jsonWriter.writeValueAsString(obj)).toString());
+        } else {
+          schemaRefNode = jsonWriter.readTree(new JSONObject(schemaFileSrc).toString());
```
manick02: Will it lead to case where schemaFileDef class is neither json nor yaml? what would be the behavior we want to exhibit in that case?

api/src/main/resources/schema/workflow.json
manick02: @tsurdilo just confirming, hope u looked at this part "existingJavaType": "io.serverlessworkflow.api.workflow.DataInputSchema"

tsurdilo: thanks, will look

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
ricardozanini: I don't think we accept a format other than json nor yaml.

tsurdilo: idk if inline would make sense for yaml defs

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
VaniHaripriya: @manick02 This scenario is handled by throwing the below exception-
"Unable to load schema defs reference file"
And in this case we are assigning null value to schema and failOnValidationErrors as true and returning the dataInputSchema.

manick02:
@tsurdilo this PR looks good from my side

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,44 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
-      dataInputSchema.setSchema(node.asText());
-      dataInputSchema.setFailOnValidationErrors(true); // default
-
-      return dataInputSchema;
-    } else {
-      dataInputSchema.setSchema(node.get("schema").asText());
+    if (node.isObject() && node.get("schema").isObject() && !node.get("schema").isEmpty()) {
+      schemaDefinition = node.get("schema");
       dataInputSchema.setFailOnValidationErrors(node.get("failOnValidationErrors").asBoolean());
+    } else {
```
fjtirado, line 58: hmmm, 
why are we not supporting this?
```
"dataInputSchema": {
      "title": "MyJSONSchema",
      "properties":{
        "firstName":{
          "type": "string"
        },
        "lastName":{
          "type": "string"
        }
      }
    }
```
In other words, why are requiring the inlined schema to be inside an "schema" field?

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,44 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
+      String schemaFileDef = node.isObject() ? node.get("schema").asText() : node.asText();
+      dataInputSchema.setFailOnValidationErrors(true);
+      dataInputSchema.setRefValue(schemaFileDef);
+      String schemaFileContent = Utils.getResourceFileAsString(schemaFileDef);
```
fjtirado: Although not related with this PR, the attempt to load the content of the schema from the URI is only partially implemented (because it does not support file or http, just classpath) and I think can be skipped. 
We can just say that is responsability of the implementation to load the schema from the uri if the refValue is not null. So if refvalue is not null, schema content must be null. 
This way we simplify the whole thing, either we have refvalue set to an uri, either we have the content read from the inlined schema. Right now, we might have both fields initialized (if the ref is an uri included in the classpath)

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,44 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
+      JsonNode schemaRefNode;
+      ObjectMapper jsonWriter = new ObjectMapper();
```
fjtirado: this is only used if the check in line 70 is true, so it should be moved withint that scope

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
```diff
@@ -46,17 +53,44 @@ public DataInputSchema deserialize(JsonParser jp, DeserializationContext ctxt)
...
+          ObjectMapper yamlReader = new ObjectMapper(new YAMLFactory());
+          Object schemaObj = yamlReader.readValue(schemaFileContent, Object.class);
+          schemaRefNode =
+              jsonWriter.readTree(
+                  new JSONObject(jsonWriter.writeValueAsString(schemaObj)).toString());
+        } else {
+          schemaRefNode = jsonWriter.readTree(new JSONObject(schemaFileContent).toString());
+        }
+
+        JsonNode schemaRef = schemaRefNode.get("schema");
```
fjtirado: hmmmm, Im not sure this is right 
schemaRefNode is suppose to have the content of the file loaded in memory, which is something like this
```
{
    "$schema": "http://json-schema.org/draft-04/schema#",
    "title": "Expression",
    "description": "Schema for expression test",
    "type": "object",
    "properties": {
        "numbers": {
            "description": "The array of numbers to be operated with",
            "type": "array",
            "items" : {
                "type" : "object",
                "properties" : {
                   "x" : {"type":"number"},
                   "y" : {"type":"number"}
                }
             }
        }
    },
    "required": ["numbers"]
}
```
so there is not "schema" field there, and I suspect the schemadefinition is always set to null when schemaFileDef is not null, making this whole loading code essentially dead code (which as I mention in another comment, should be removed)
You should be able to verify what Im saying through testing

fjtirado (commented):
@VaniHaripriya 
Please take a look at my comments. 
I believe the part that try to load the schemaRef to set the schemaDef is not working properly and I think we need to discuss if it makes sense at all to try to load it. 
I vote for simplification and states in the java doc that if schemaRef is not null (the content is expected to be into an uri)  schemaDef will be null (the content is not inlined) and the other way around, if schemaDef is not null (it means the schema is inlined) schemaRef will be null (it makes no sense to have an uri if the schema is inlines)
We can state that without breaking compatibility because as I tried to proof in my comments, that loading code is not working even for the only case is expected to work (a classpath URI)

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
ricardozanini: I agree since we are doing the same for the OpenAPI, Async, and gRPC functions (not loading the resources). We can have an implementation later that would fetch from classpath or other sources and have the possibility to let the runtime to override this behavior if needed.

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
ricardozanini: This should be changed in the spec: https://github.com/serverlessworkflow/specification/blob/main/schema/workflow.json#L48-L74

The spec requires the `schema` attribute. But yeah, I think it would be an excellent enhancement to the spec. Can you open an issue there?

tsurdilo: this would maybe work with short schemas but would be a pain for anything concrete imo. in most cases offloading schema to separate file so it can be validated and reused would be prefered

VaniHaripriya:
@fjtirado Could you please verify the changes and let me know if I need to do any other changes or we are good with this PR.

tsurdilo:
Whats the overall status of this pr? Will there be more updates/changes?

api/src/main/java/io/serverlessworkflow/api/deserializers/DataInputSchemaDeserializer.java
fjtirado: @tsurdilo 
After this PR we will be supporting this
```
"dataInputSchema": {
      "schema": {"title": "MyJSONSchema",
      "properties":{
        "firstName":{
          "type": "string"
        },
        "lastName":{
          "type": "string"
        }
      }
}
    }
```
and 
```
"dataInputSchema": {
  "filecontainingschema.json"
}
```
and
```
"dataInputSchema": {
  "schema": "filecontainingschema.json"
}
```

I think spec should be changed to also support this

```
"dataInputSchema": {
 {"title": "MyJSONSchema",
      "properties":{
        "firstName":{
          "type": "string"
        },
        "lastName":{
          "type": "string"
        }
      }
    }
```
There is an asymetry there., that should be corrected. A minor thing that is slightly related with this PR and should be done with a different one if we agree. Lets merge this one as it is

fjtirado (approved):
Great work @VaniHaripriya

VaniHaripriya:
> Whats the overall status of this pr? Will there be more updates/changes?

@tsurdilo  There won't be any updates/changes on this PR. Thank you.

ricardozanini (approved):

VaniHaripriya:
@tsurdilo Could you please take a look at this PR and merge it if every thing looks fine.

tsurdilo (approved):

PR #210, opened 2022-07-14
[wip] Change schema loader to get schemas from website

tsurdilo:
Still need to add more tests
Note that validation errors are json format now, users would need to read the json and display to user

### DIFF
--- a/api/src/main/java/io/serverlessworkflow/api/validation/WorkflowSchemaLoader.java
+++ b/api/src/main/java/io/serverlessworkflow/api/validation/WorkflowSchemaLoader.java
@@ -15,26 +15,50 @@
  */
 package io.serverlessworkflow.api.validation;
 
-import io.serverlessworkflow.api.schemaclient.ResourceSchemaClient;
+import java.io.*;
+import java.net.URL;
+import java.nio.charset.Charset;
 import org.everit.json.schema.Schema;
 import org.everit.json.schema.loader.SchemaLoader;
-import org.everit.json.schema.loader.internal.DefaultSchemaClient;
+import org.json.JSONException;
 import org.json.JSONObject;
-import org.json.JSONTokener;
 
 public class WorkflowSchemaLoader {
-  private static final JSONObject workflowSchema =
-      new JSONObject(
-          new JSONTokener(WorkflowSchemaLoader.class.getResourceAsStream("/schema/workflow.json")));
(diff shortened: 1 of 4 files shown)


api/src/main/java/io/serverlessworkflow/api/validation/WorkflowSchemaLoader.java
```diff
@@ -15,26 +15,50 @@
...
+import org.json.JSONException;
 import org.json.JSONObject;
-import org.json.JSONTokener;
 
 public class WorkflowSchemaLoader {
-  private static final JSONObject workflowSchema =
-      new JSONObject(
-          new JSONTokener(WorkflowSchemaLoader.class.getResourceAsStream("/schema/workflow.json")));
+  // note update this when supporting new version
+  private static final String BASE_SCHEMA_URL = "https://serverlessworkflow.io/schemas/0.8/";
```
manick02, line 28: I think we have to paramterise version part of https://serverlessworkflow.io/schemas/0.8 (leave out 0.8 from base schema).  Would it not be bit Odd for users to download the schema from a not well known Url? It adds bit of security concern as well. Sorry, my initial thoughts on this

tsurdilo: thanks for input, what would you recommend? we host schemas on website a d thought we can use it, u think having them as part of sdk and reading local would be better?

manick02: IMO, if there is a way we can publish the schema as a bundle in a mvn repository or somewhere and have a way to include them as part of build process may be better. Somethink like described here [https://stackoverflow.com/questions/51423201/use-maven-repo-to-distribute-json-schema-files](https://stackoverflow.com/questions/51423201/use-maven-repo-to-distribute-json-schema-files). I honestly dont know how this would translate for us.  For some environments, getting the schema from the website may be fine, but for some bundling it as part of build process could be better. I am trying to see if this pattern of having schema in a website and later on lazy loading would be fine for most of the cases bit hard to reason out to be honest

tsurdilo: love the idea, ty! will try to impl

PR #228, opened 2023-05-08
Key property is missing in the Workflow

spolti:
fixes https://github.com/serverlessworkflow/sdk-java/issues/227

**Many thanks for submitting your Pull Request :heart:!**

**What this PR does / why we need it**:

**Special notes for reviewers**:

**Additional information (if needed):**

PR #234, opened 2023-07-04
[4.0.x] Key property is missing in the Workflow

ricardozanini:
Cherry-pick for #228


tsurdilo:
i released 4.0.4 before seeing this sorry, will include it in 4.0.5.Final

PR #238, opened 2023-07-06
Removing gson and everit dependencies

fjtirado:

It removed org.json and everit validation dependencies and add json-schema-validator for validation.
See https://github.com/everit-org/json-schema#when-to-use-this-library
Since we are using Jackson for serialization/deserialization of the workflow, we should use https://github.com/networknt/json-schema-validator instead of everit, and hence avoid having dependencies to two different json parsers. 

This remove unneeded dependencies for poms and slightly change the validator interface (it now returns a Collection of ValidationErrors rather than a List, which easily remove duplicated validation messages) 


ricardozanini:
Although this is very important, we need to handle the CVEs while they don't release a new version:

https://mvnrepository.com/artifact/com.github.java-json-tools/json-schema-validator/2.2.14

1. Upgrade Guava - [CVE-2023-2976](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-2976) and [CVE-2020-8908](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908)
2. Upgrade testng-core - [CVE-2022-4065](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-4065)

pom.xml
```diff
@@ -112,6 +112,11 @@
                 <artifactId>jackson-databind</artifactId>
                 <version>${version.com.fasterxml.jackson}</version>
             </dependency>
+            <dependency>
+              <groupId>com.github.java-json-tools</groupId>
+              <artifactId>json-schema-validator</artifactId>
+              <version>${version.com.github.java-json-tools}</version>
```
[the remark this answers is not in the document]
fjtirado: Working on fixing vulnerabilities of the jackson base library

fjtirado:
> Although this is very important, we need to handle the CVEs while they don't release a new version:
> 
> https://mvnrepository.com/artifact/com.github.java-json-tools/json-schema-validator/2.2.14
> 
>     1. Upgrade Guava - [CVE-2023-2976](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2023-2976) and [CVE-2020-8908](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2020-8908)
> 
>     2. Upgrade testng-core - [CVE-2022-4065](https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2022-4065)

To remove this CVEs, lets use https://github.com/networknt/json-schema-validator, reworking the PR

tsurdilo:
Will review. Just note that if you add this much refactoring would be nice to add some additional tests for it in addition to existing.

api/src/main/java/io/serverlessworkflow/api/deserializers/RetriesDeserializer.java
```diff
@@ -69,21 +67,10 @@ public Retries deserialize(JsonParser jp, DeserializationContext ctxt) throws IO
     } else {
       String retriesFileDef = node.asText();
       String retriesFileSrc = Utils.getResourceFileAsString(retriesFileDef);
-      JsonNode retriesRefNode;
-      ObjectMapper jsonWriter = new ObjectMapper();
+      ;
```
tsurdilo: extra?

fjtirado: good catch!!

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
```diff
@@ -61,298 +62,273 @@ public WorkflowValidator setSource(String source) {
...
-                          && !m.equals("#/retries: expected type: JSONObject, found: JSONArray"))) {
-                        addValidationError(m, ValidationError.SCHEMA_VALIDATION);
-                      }
-                    });
-          }
-        }
-      } catch (Exception e) {
-        logger.error("Schema validation exception: " + e.getMessage());
+
+        JsonSchemaFactory.getInstance(VersionFlag.V202012)
```
ricardozanini: Isn't [V7](https://github.com/networknt/json-schema-validator/blob/master/src/main/java/com/networknt/schema/SpecVersion.java#L26) the most used? Maybe we should use the same one defined in the old dependency?

fjtirado: V7 is fine, I believe old one was V2, but was unsure about it

ricardozanini (approved):

tsurdilo:
thanks for the updates. i would really like to request some more tests since its a bigger refactor. can we add some positive/negative tests (and not just one please :) )

fjtirado:
> thanks for the updates. i would really like to request some more tests since its a bigger refactor. can we add some positive/negative tests (and not just one please :) )

@tsurdilo Since this PR, despite apparently large, is not adding any new functionality (just keeping the existing one using a library that avoid dependencies that bring vulnerabilities), fortunately (kudos for the original writer ;)), the tests that are related with the change: basically serialization/deserialization and validation, were already there. In fact, they were changed as result of this change because the valitadion error messaged are library dependent.

api/src/main/java/io/serverlessworkflow/api/mapper/JsonObjectMapperFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.api.mapper;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class JsonObjectMapperFactory {
```
tsurdilo, line 20: Not sure i get point of this class. Just maybe have static instance of yaml/json mappers in a util class?

api/src/main/java/io/serverlessworkflow/api/mapper/YamlObjectMapperFactory.java
```diff
@@ -0,0 +1,27 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.api.mapper;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+
+public class YamlObjectMapperFactory {
```
tsurdilo, line 20: Same here

api/src/main/java/io/serverlessworkflow/api/validation/ValidationError.java
```diff
@@ -43,4 +45,18 @@ public void setType(String type) {
   public String toString() {
     return String.format(MSG_FORMAT, type, message);
   }
+
+  @Override
+  public int hashCode() {
```
tsurdilo: is hashcode and equals override here needed or just cosmetic?

api/src/main/resources/schema/workflow.json
```diff
@@ -1,5 +1,5 @@
 {
-  "$id": "https://wg-serverless.org/workflow.schema.json",
+  "$id": "classpath:schema/workflow.schema.json",
```
tsurdilo, line 2: why is this change needed?

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
```diff
@@ -27,24 +28,24 @@
...
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 public class WorkflowValidatorImpl implements WorkflowValidator {
 
   private static final Logger logger = LoggerFactory.getLogger(WorkflowValidatorImpl.class);
   private boolean schemaValidationEnabled = true;
-  private List<ValidationError> validationErrors = new ArrayList<>();
-  private Schema workflowSchema = WorkflowSchemaLoader.getWorkflowSchema();
+  private Collection<ValidationError> validationErrors = new LinkedHashSet<>();
```
tsurdilo: Do you know if this collection is ordered? If its not then I dont think we wanna do this change
as we need to preserve the order of errors generated when iterating over this collection. List is always ordered.

(I dont think its possible to know if collection is gonna be ordered here but let me know if im missing something)

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
```diff
@@ -61,298 +62,272 @@ public WorkflowValidator setSource(String source) {
...
-      if (workflow == null) {
-        workflow = Workflow.fromSource(source);
-      }
+    } else if (workflow == null) {
+      workflow = Workflow.fromSource(source);
+    }
 
-      List<FunctionDefinition> functions =
-          workflow.getFunctions() != null ? workflow.getFunctions().getFunctionDefs() : null;
+    List<FunctionDefinition> functions =
```
tsurdilo, line 87: Can you check on these formatting changes? Iirc we have auto-formatting built in.

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
fjtirado: List are always ordered, but  I think the requirement here is that errors are sorted by insertion order and this is also provided by LinkedHashSet (and not necessarily by List, because implementations can always set the position of indexes in a different way than insertion order)  
The reason I changed to Collection is because it allows using a Set implementation, which prevents duplication, which is something that I think should also be a requirement for validations errors. To prevent duplication, I added hashcode and equals methods to ValidationError class.  
I think the discussion should be if we change to Collection or directly to Set, because declaring validation errors as Set ensures there is not duplication (and we LinkedHashSet you can keep the previous requirement of having errors sorted by order of appearance)  
To stress my point, I want to note that the validation library API returns a set with the validation errors.

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
fjtirado: Yes, I will review this, thanks

api/src/main/java/io/serverlessworkflow/api/mapper/JsonObjectMapperFactory.java
fjtirado: This avoid creating a new object mapper instance every time is needed

api/src/main/java/io/serverlessworkflow/api/mapper/YamlObjectMapperFactory.java
fjtirado: same, to avoid creating multiple object mapper instances

api/src/main/java/io/serverlessworkflow/api/validation/ValidationError.java
fjtirado: This is changed to avoid duplicates when inserting validation errors into a Set implementation

api/src/main/resources/schema/workflow.json
fjtirado: the new validation library tries to download the schema from a not existing url (Because wg-serverless-org is fake), hence changing to real location (which is library classpath)

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
fjtirado: I reviewed and this is performed by the maven validator, I think this class was not formatted previously, to properly review it, please check hide whitespace checkbox in config

api/src/main/java/io/serverlessworkflow/api/mapper/JsonObjectMapperFactory.java
tsurdilo: ok got it. i think since you return just the basic "new" mapper just define it in static field in some util class no need imo
for separate class. if you want to allow users to set specific configuration for the mappers then having it in some separate class like this maybe makes more sense

api/src/main/java/io/serverlessworkflow/api/validation/ValidationError.java
tsurdilo: do you have a test that makes sure this is done correctly? what if you have multiple workflow states in definition
and both produce the same error? trying to understand why duplication is a bad thing

api/src/main/resources/schema/workflow.json
tsurdilo: tell me if im wrong you already define the hard-coded path to the schema in code, if so then why here? i think it was by design to put some id like this and not use classpath

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
tsurdilo: ok yeah im not sure this duplication requirement is valid one but will think more about it (per comment above also please let me know if we have a test for this possible duplication)
either way i think that having iteration order is as much important as possibly this duplication issue if that is indeed an issue

api/src/main/java/io/serverlessworkflow/api/mapper/JsonObjectMapperFactory.java
fjtirado: This is done in a separate class to only initialize the mapper object when it is really used. If you put it into an utils class it will get initialized the firts time you use the utilities class, even in the utility method being used is unrelated (imagine an use that never use yaml format)

api/src/main/java/io/serverlessworkflow/api/validation/ValidationError.java
fjtirado: Its pretty trivial IDE generated code, but I will add the tests

api/src/main/resources/schema/workflow.json
fjtirado: this id path is used to upload other schemas referred from this schema
So, if we leave wg-.serverless.org, the library will fail to retrieve, any nested shema, for example, start.json.

api/src/main/java/io/serverlessworkflow/api/validation/ValidationError.java
fjtirado: Added

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
ricardozanini: @fjtirado we could use `Set` instead of `Collection` to describe the interface intention clearly. 
@tsurdilo I don't think we would like a duplicated entry of a validation error. In which scenario would this fit?

api/src/test/java/io/serverlessworkflow/api/test/ValidationErrorTest.java
```diff
@@ -0,0 +1,53 @@
...
+
+    assertEquals(duplicatedErrors.size(), 3);
+    assertEquals(errors.size(), 2);
+
+    assertEquals(duplicatedErrors.iterator().next().getMessage(), "This is the first message");
+    assertEquals(errors.iterator().next().getMessage(), "This is the first message");
+  }
+
+  private Collection<ValidationError> addMessagesToCollection(Collection<ValidationError> errors) {
+    ValidationError first = new ValidationError();
```
tsurdilo: i dont really understand the purpose of this test. in a workflow definition you can have multiple workflow states create same error message. for example you have two states that are missing the name prop or something. 
in this case imho it does not make sense to treat them as same validation error and remove one. maybe the check on uniqueness of a validation error should be the workflow state itself, but even then i dont understand where in validation a multiple validation error for same workflow state can be produced?

ricardozanini: This is a valid use case. But I believe that the validation error will be different between the states, no?

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
tsurdilo: @ricardozanini different workflow states could create same validation error. i think current impl in this pr checks on type and message for comparison, which imho is not correct: https://github.com/serverlessworkflow/sdk-java/pull/238/files#diff-3f4d6094e04aefe8a2e00aaf6354e106bdd4411a996cff4d5126b7ac56307a5eR60

if you really want to enforce some sort of uniqueness imho then do it per wfstate+type+message, but again i dont understand why this is a concern? So imho we should just leave it as List as before we have 100% iteration order 
and don't worry about this added uniqueness constraint.

api/src/test/java/io/serverlessworkflow/api/test/ValidationErrorTest.java
tsurdilo: i dont think it will be different, for custom validation errors we can generate same message for different states i think

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
ricardozanini: hmm I see, there's no way in the current `ValidationError` class to distinguish between states. So yes, in this case, we should use a `List` instead. Or this validation error report is per state? I would have to try it. One way of testing it is to run the test in a SW with the same error in different states.

api/src/test/java/io/serverlessworkflow/api/test/ValidationErrorTest.java
tsurdilo: i think this test just shows that java Collection class works :) not sure it really shows that during validation our validation CAN produce multiple same validation errors and then show the benefit of the added change

tsurdilo:
I think the validation piece (List/Collection + the added imho ambiguous uniqueness constraint) is whats holding up the pr
I dont have issue making the library change, but the pr also adds this other piece.
Would help maybe if you can remove that added uniques constraint and only have in PR the lib change stuff and then we can update this and discuss that uniqueness stuff in separate pr so we don't hold up this one?

api/src/test/java/io/serverlessworkflow/api/test/ValidationErrorTest.java
fjtirado: Since the list vs set discussion is not really relevant to this PR (I made the change because I believe it was self evident and because one test was failing in validation test with previous library in this same pr) I will rollback to List 
I honestly cannot see the point of adding exactly the same error meessage twice in a collection of validation errors, but I do not think that discussion should defer this PR, which target is to remove unneeded dependencies.

ricardozanini: @fjtirado I believe the collection of validation errors as a whole (in the flow context), shouldn't have a duplication. I would need to test on my end to see how that pans out. If this collection is per state/context, then it should avoid duplications. Otherwise, if it's part of the whole context, it won't. But then, how to differentiate between the states? Anyhow, depends on the context. I'd have to try/write a test case to verify how the SDK behaves.

validation/src/main/java/io/serverlessworkflow/validation/WorkflowValidatorImpl.java
tsurdilo: agree. i think having a test that generates same custom validation error on 2+ states and then see if this uniqueness kicks in and does not remove the error that actually is valid (so not duplicate but same for a different wf state) would be nice

fjtirado: Reverted to List

ricardozanini (approved):

ricardozanini:
@tsurdilo I think this PR is ready since the conversation about `List` or `Set` has been settled down and the commit reverted back to what it was?

cdavernas (approved):

PR #249, opened 2023-07-28
Introduce Release Workflow

ricardozanini:

With this PR we can automate the release.

Work to be done:

- [x] - Add the GPG_PRIVATE_KEY and GPG_PASSPHRASE secrets to this repo
- [x] - Add the `.github/release/maven-settings.xml.gpg` file signed with the GPG key

I've removed the current actions since they were pushing SNAPSHOTs to Maven Central and not doing an actual release.

To do a release, one must open a PR and change the info in `./github/project.yml`. The action will branch the repo, and push the tags and the released artifact to Maven Central.


ricardozanini:
@tsurdilo we just need the keys configured as secrets in this repo, please see the PR description 🙏

ricardozanini:
@manick02 @fjtirado since you're the latest contributors, do you mind taking a look too?

cdavernas (approved):

manick02:
@ricardozanini Thanks for tagging me. I am not aware of how the current release process works. Currently fixes for older sdk versions are built out from the branches 4.0, 3.0 etc. These release workflow changes I assume would be propagated to the branches from which older sdks are supported?

ricardozanini:
@manick02 I'll cherry-pick to 4.0.x branch if it makes sense there. This workflow tags the main branch upon release with the name version. About branching, I'll continue doing it manually from the tag after the release.

validation/pom.xml
```diff
@@ -124,7 +124,7 @@
                 </executions>
             </plugin>
             <plugin>
-                <groupId>com.coveo</groupId>
+                <groupId>com.spotify.fmt</groupId>
```
manick02, line 127: nit: should we have made this change in different PR?

manick02 (approved):
Looks good to me

validation/pom.xml
ricardozanini: The old one had conflicts during the build and it was outdated :(

PR #279, opened 2023-10-25
Keep id as mandatory for backward compatibility.

fjtirado:


ricardozanini:
@fjtirado can you please sign your commit?

ricardozanini (approved):

cdavernas (approved):

ricardozanini:
This change is needed since without it, it will break the API interface for a minor release.

PR #284, opened 2023-10-27
Force id to be mandatory

fjtirado:

This makes id mandatory, keeping current constructor unchanged from the previous version

SDK has currently two misaligments with the spec: id and name are mandatory, while the former is mandatory only if there is no key and the latter is optional. Since making id optionally depending on key is not a good idea that will be probably changed in 0.9 (where ID will be mandatory always and key removed or make optional), I think is better to postpone the API breaking change to the migration to 0.9

### DIFF
--- a/api/src/main/resources/schema/workflow.json
+++ b/api/src/main/resources/schema/workflow.json
@@ -161,6 +161,7 @@
     }
   },
  "required": [
+        "id",
         "name",
         "version",
         "states"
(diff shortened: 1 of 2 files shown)


ricardozanini:
This PR partially reverts what was done in this one:

https://github.com/serverlessworkflow/sdk-java/commit/8fbc3bf1dfbdd362ccf54b77b0b7f08aec29d6c9

The author in that commit, by adding `key`, enforced the spec. This is fine, but broke the API contract since the json schema generator is now different not having the `id` as a parameter.

Originally, the SDK forced the `id` to be required. We can review this in another major version/step. Also see: https://github.com/serverlessworkflow/specification/issues/742

ricardozanini (approved):

ricardozanini:
@manick02 can you take a look too please?

manick02 (approved):
Looks good to me

cdavernas (approved):

PR #413, opened 2024-08-22
Add single value enum when using const keyword for compatiblity

uasouz:
Use extra single value enum where const is used due to lack of support from jsonschema2pojo

jsonschema2pojo does not support `const` keyword yet.
This was causing a bug when using resusable tasks that have no value on `with` block. It would always match to `CallAsyncAPI`.

In order to achieve proper behaviour I added a single value `enum` below each `const`. This will make it compatible with jsonschema2pojo while they solve the issue #1001.

https://github.com/joelittlejohn/jsonschema2pojo/issues/1001.

Also addded a test to validate the behaviour

I think this may be a good idead to adopt as a single enum on the main spec too. const and single value enum are the same thing under the hood, for jsonschema2pojo at least.

### DIFF
--- a/api/src/main/resources/schema/workflow.yaml
+++ b/api/src/main/resources/schema/workflow.yaml
@@ -202,6 +202,7 @@ $defs:
           call:
             type: string
             const: asyncapi
+            enum: [asyncapi]
           with:
             type: object
             title: AsyncApiArguments
@@ -247,6 +248,7 @@ $defs:
           call:
             type: string
             const: grpc
+            enum: [grpc]
           with:
             type: object
             title: GRPCArguments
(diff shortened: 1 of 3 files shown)


fjtirado:
@uasouz I think we should add support to const. 
To implement this workaround, we should change the original schema in the spec repo (the main target is for them to not differ in a single comma, actually copying the schema was a workaroung till pojo2scheme add support for non local file paths)

uasouz:
@fjtirado 100% agree, we should.
I will then open another PR on the main specification repository for the workaround.
So if it is ok, you can close this PR.

fjtirado:
@uasouz Actually I believe we should use another PR to add support to const. The idea will be to incorporate to our serializer the const check.

fjtirado:
@ricardozanini Please close this PR and assign https://github.com/serverlessworkflow/sdk-java/issues/418 to me

PR #419, opened 2024-08-28
[Fix #418] Adding const support

### LINKED ISSUE #418: Support const keyword in generator
Currently the const word is ignored

**Why is this needed**:
To distinghish, in CallTask, the right class when there is no param.

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/418


fjtirado:
Pattern annotation has been added to generated POJO. 
Next step is to make jackson deserializer validate JSR 303 annotations

ricardozanini (approved):

Issue #449, opened 2024-10-28
Add pattern  and format check for Strings in one/any/allOf 


fjtirado:
String patterns are currently ignored for strings that are an option in anyOf/allOf/oneOf construct

This means that, for example, ErrorInstance section of the schema

```
 instance:
        title: ErrorInstance
        description: A JSON Pointer used to reference the component the error originates from.
        oneOf:
          - title: LiteralErrorInstance
            description: The literal error instance.
            type: string
            format: json-pointer
          - $ref: '#/$defs/runtimeExpression'
            title: ExpressionErrorInstance
            description: An expression based error instance.
```

is currently translated to class

```
public class ErrorInstance
    implements OneOfValueProvider
{

    private Object value;
    private String literalErrorInstance;
    private String expressionErrorInstance;

    public ErrorInstance(String literalErrorInstance) {
        this.value = literalErrorInstance;
        this.literalErrorInstance = literalErrorInstance;
        this.value = literalErrorInstance;
        this.expressionErrorInstance = literalErrorInstance;
    }

    public ErrorInstance() {
    }

    @Override
    public Object get() {
        return value;
    }

    public String getLiteralErrorInstance() {
        return literalErrorInstance;
    }

    public String getExpressionErrorInstance() {
        return expressionErrorInstance;
    }

}

```

The generated  constructor should be
```
    public ErrorInstance(String erroString) {

        if (fulfillsRuntimeExpressionPattern(errorString)) {
               this.expressionErrorInstance = errorString;
        } else { 
              this.literalErrorInstance = errorString;
       }
        this.value = errorString;
    }

```

PR #451, opened 2024-11-04
[Fix_#449] Checking pattern

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/449

### DIFF
--- a/custom-generator/src/main/java/io/serverlessworkflow/generator/AllAnyOneOfSchemaRule.java
+++ b/custom-generator/src/main/java/io/serverlessworkflow/generator/AllAnyOneOfSchemaRule.java
@@ -23,6 +23,7 @@ import com.sun.codemodel.JBlock;
 import com.sun.codemodel.JClass;
 import com.sun.codemodel.JClassAlreadyExistsException;
 import com.sun.codemodel.JClassContainer;
+import com.sun.codemodel.JConditional;
 import com.sun.codemodel.JDefinedClass;
 import com.sun.codemodel.JExpr;
 import com.sun.codemodel.JFieldVar;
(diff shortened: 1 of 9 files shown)


custom-generator/src/main/java/io/serverlessworkflow/generator/AllAnyOneOfSchemaRule.java
```diff
@@ -54,7 +59,35 @@ class AllAnyOneOfSchemaRule extends SchemaRule {
     this.ruleFactory = ruleFactory;
   }
 
-  private static class JTypeWrapper {
+  private static final String REF = "$ref";
+  private static final String PATTERN = "pattern";
+
+  private enum Format {
+    URI_TEMPLATE("^[A-Za-z][A-Za-z0-9+\\-.]*://.*");
```
fjtirado, line 66: The idea is to add adittional formats when needed. 
Right now, the only format that is used by the spec and not handle is json-pointer, but I could not find a good one for this ABNF yet
```
  json-pointer    = *( "/" reference-token )
      reference-token = *( unescaped / escaped )
      unescaped       = %x00-2E / %x30-7D / %x7F-10FFFF
         ; %x2F ('/') and %x7E ('~') are excluded from 'unescaped'
      escaped         = "~" ( "0" / "1" )
        ; representing '~' and '/', respectively
```

impl/src/main/java/io/serverlessworkflow/impl/DefaultTaskExecutorFactory.java
```diff
@@ -0,0 +1,42 @@
...
+import io.serverlessworkflow.api.types.Task;
+import io.serverlessworkflow.api.types.TaskBase;
+
+public class DefaultTaskExecutorFactory implements TaskExecutorFactory {
+
+  protected DefaultTaskExecutorFactory() {}
+
+  private static TaskExecutorFactory instance = new DefaultTaskExecutorFactory();
+
+  public static TaskExecutorFactory get() {
```
fjtirado, line 28: This has been made public, allowing other implementations to reuse the code executing task as desired.

ricardozanini (approved):

PR #457, opened 2024-11-04
Release 7.0.0-alpha5

fjtirado:

### DIFF
--- a/.github/project.yml
+++ b/.github/project.yml
@@ -1,3 +1,3 @@
 release:
-  current-version: 7.0.0-alpha2
+  current-version: 7.0.0-alpha5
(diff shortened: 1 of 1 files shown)


.github/project.yml
```diff
@@ -1,3 +1,3 @@
 release:
-  current-version: 7.0.0-alpha2
+  current-version: 7.0.0-alpha5
```
ricardozanini, line 2:
```suggestion
  current-version: 7.0.0-alpha3
```

We don't have 3 and 4 🤔

fjtirado: yes, but the one in the schema is the alpha5, I think is better to align with it, isnt it?

ricardozanini (approved):

PR #525, opened 2025-01-31
Releasing 7.0.0

fjtirado:


.github/project.yml
```diff
@@ -1,3 +1,3 @@
 release:
-  current-version: 7.0.0-alpha5.1
+  current-version: 7.0.0
   next-version: 7.0.0-SNAPSHOT
```
ricardozanini:
```suggestion
  next-version: 8.0.0-SNAPSHOT
```

ricardozanini (changes requested):
We can't merge until we fix the SNAPSHOT version.

fjtirado:
> We can't merge until we fix the SNAPSHOT version.

You mean just here, or we need need to change something somewhere else?

fjtirado:
@ricardozanini Im thinking that probably it would  be better to keep 7.0.0-SNAPSHOT rather than 8.0.0-SNAPSTHOT. It is very likely that we release 7.0.1 as long as the reference implementation is including more features (and the dsl schema will probably get minor updates too)
wdyt?

ricardozanini:
@fjtirado this is fine. Next time, you just update the `version: 7.0.1`. Next version means the major one.

ricardozanini (approved):

PR #544, opened 2025-03-04
fixes #543 change hibernate-validator and expressly to test scope

### LINKED ISSUE #543: change hibernate-validator and expressly to test scope
Those two dependencies are not needed in runtime and they are only used by the test cases in api and impl modules. 

**Why is this needed**:
By moving the two dependencies to the test scope, we can minimize the dependency for api and impl modules. 

If this is agreed, I can submit a PR to resolve it. Thanks a lot for providing the Java sdk and reference implementation.

stevehu:

### DIFF
--- a/api/pom.xml
+++ b/api/pom.xml
@@ -37,15 +37,17 @@
             <groupId>jakarta.validation</groupId>
             <artifactId>jakarta.validation-api</artifactId>
         </dependency>
-        <dependency>
-          <groupId>org.hibernate.validator</groupId>
-          <artifactId>hibernate-validator</artifactId>
-        </dependency>
-        <dependency>
-          <groupId>org.glassfish.expressly</groupId>
-          <artifactId>expressly</artifactId>
-        </dependency>
         <!-- test -->
+        <dependency>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <scope>test</scope>
+        </dependency>
+        <dependency>
+            <groupId>org.glassfish.expressly</groupId>
+            <artifactId>expressly</artifactId>
+            <scope>test</scope>
+        </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
             <artifactId>junit-jupiter-api</artifactId>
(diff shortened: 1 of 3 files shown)


api/pom.xml
```diff
@@ -37,15 +37,17 @@
             <groupId>jakarta.validation</groupId>
             <artifactId>jakarta.validation-api</artifactId>
         </dependency>
+        <!-- test -->
         <dependency>
-          <groupId>org.hibernate.validator</groupId>
-          <artifactId>hibernate-validator</artifactId>
+            <groupId>org.hibernate.validator</groupId>
+            <artifactId>hibernate-validator</artifactId>
+            <scope>test</scope>
```
fjtirado, line 44: Actually these are neeeed at runtime, not just for test, so it should be runtime not test

api/pom.xml
```diff
@@ -37,15 +37,17 @@
...
         </dependency>
         <dependency>
-          <groupId>org.glassfish.expressly</groupId>
-          <artifactId>expressly</artifactId>
+            <groupId>org.glassfish.expressly</groupId>
+            <artifactId>expressly</artifactId>
+            <scope>test</scope>
```
fjtirado, line 49: Same here

impl/core/pom.xml
```diff
@@ -34,6 +34,16 @@
             <groupId>net.thisptr</groupId>
             <artifactId>jackson-jq</artifactId>
         </dependency>
+        <dependency>
```
fjtirado, line 37: Once you made the other runtime, this does not need to be added and this file should remain unchanged

impl/http/pom.xml
```diff
@@ -21,6 +21,16 @@
             <groupId>io.serverlessworkflow</groupId>
             <artifactId>serverlessworkflow-impl-core</artifactId>
         </dependency>
+        <dependency>
```
fjtirado, line 24: Same, not needed once you set the dependency to runtime, this file should not be modified

fjtirado (changes requested):
These two dependencies are needed for the SDK to work on runtime, so they should be changed to runtime not to test
That way, you do not need to change the impl poms

api/pom.xml
fjtirado, line 44:
```suggestion
            <scope>runtime</scope>
```

api/pom.xml
fjtirado, line 49:
```suggestion
            <scope>runtime</scope>
```

PR #560, opened 2025-04-25
Update produceevent.json

fjtirado:

### DIFF
--- a/api/src/main/resources/schema/produce/produceevent.json
+++ b/api/src/main/resources/schema/produce/produceevent.json
@@ -8,7 +8,8 @@
       "minLength": 1
     },
     "data": {
-      "type": "string",
+      "type": "object",
+      "existingJavaType": "com.fasterxml.jackson.databind.JsonNode",
       "description": "Workflow expression which selects parts of the states data output to become the data of the produced event"
     },
     "contextAttributes": {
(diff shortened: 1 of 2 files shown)


api/src/main/resources/schema/produce/produceevent.json
```diff
@@ -8,7 +8,8 @@
       "minLength": 1
     },
     "data": {
-      "type": "string",
+      "type": "object",
```
masayag, line 11: According to the spec this can be either a string or an Object.
If this shouldn't be a string, can you update the description to reflect the change?

> If object type, a custom object to become the data (payload) of produced event.


https://github.com/serverlessworkflow/specification/blob/0.8.x/specification.md#producedevent-definition

masayag: perhaps this description needs to be updated as well:
https://github.com/serverlessworkflow/sdk-java/pull/561/files#diff-a1593297ec9764a50a83b736ce196d5272b2b4281ae73613f10d742598638568R21

fjtirado: This schema is just an utility to make the generator work. In order the generator to work for both string and object, the generated POJO should return a JsonNode.

PR #601, opened 2025-05-27
[4.x] - Upgrade to Java 17

ricardozanini:
Since this stream is still in use, we must upgrade to Java 17 to keep it up to date with Quarkus LTS and other dependencies, or we will fall behind on CVE fixes.

Additionally, I removed a few old content.


ricardozanini:
@fjtirado @gmunozfe can you please take a look?

PR #678, opened 2025-07-29
Improved handling of WorkflowModel and Cognisphere

treblereel:


ricardozanini:
@treblereel, @fjtirado will fix the Map/Cognisphere at the engine level. Can we change the imports only?

PR #679, opened 2025-07-29
Add more unit tests and docs for fluent

ricardozanini:
Added a few more unit tests for `agentic` and proper documentation for the new `fluent` module. Introduced the new `serverlessworkflow-lambda-fluent` module to bring in Fluent + Java impl in a single dependency.


fjtirado (approved):
Lamba-fluent was added to reduce the number of dependencies to be written in the final user pom, but it is not mentioned in readme

ricardozanini:
> Lamba-fluent was added to reduce the number of dependencies to be written in the final user pom, but it is not mentioned in readme

@fjtirado this will be part of another doc. This is just for the `fluent` module.

PR #681, opened 2025-07-31
Introduce LangChain4j Agentic Workflow Implementation

ricardozanini:
Depends on https://github.com/langchain4j/langchain4j/pull/3458

In this PR:

- Add a new `serverlessworkflow-fluent-agentic-langchain4j` module to implement `langchain4j-agentic` SPI. So that we can use our workflow engine to run Agentic use cases via LC4J Workflow SPI.
- Adjust `serverlessworkflow-experimental-agentic` to handle `AgenticScope` in the workflow context
- Add a new CI to run solely Integration Tests since now we depend on Ollama to run Agentic use cases on this new module.

Later, I'll add more integration tests, adjust the CI to install Ollama, and the docs.

### DIFF
--- a/experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModel.java
+++ b/experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModel.java
@@ -15,16 +15,17 @@
  */
 package io.serverlessworkflow.impl.expressions.agentic;
 
-import dev.langchain4j.agentic.cognisphere.Cognisphere;
+import dev.langchain4j.agentic.scope.AgenticScope;
 import io.serverlessworkflow.impl.WorkflowModel;
 import io.serverlessworkflow.impl.expressions.func.JavaModel;
 import java.util.Collection;
+import java.util.Map;
 import java.util.Optional;
 
 class AgenticModel extends JavaModel {
 
-  AgenticModel(Cognisphere cognisphere) {
-    super(cognisphere);
+  AgenticModel(AgenticScope agenticScope) {
+    super(agenticScope);
   }
 
   @Override
@@ -39,8 +40,10 @@ class AgenticModel extends JavaModel {
 
   @Override
   public <T> Optional<T> as(Class<T> clazz) {
-    if (Cognisphere.class.isAssignableFrom(clazz)) {
+    if (AgenticScope.class.isAssignableFrom(clazz)) {
       return Optional.of(clazz.cast(object));
+    } else if (Map.class.isAssignableFrom(clazz)) {
+      return Optional.of(clazz.cast(((AgenticScope) object).state()));
     } else {
       return super.as(clazz);
     }
--- a/experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModel.java
+++ b/experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModel.java
@@ -16,20 +16,15 @@
 package io.serverlessworkflow.impl.expressions.agentic;
 
 import dev.langchain4j.agentic.cognisphere.Cognisphere;
-import dev.langchain4j.agentic.cognisphere.ResultWithCognisphere;
 import io.serverlessworkflow.impl.WorkflowModel;
 import io.serverlessworkflow.impl.expressions.func.JavaModel;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.Optional;
 
 class AgenticModel extends JavaModel {
 
-  private final Cognisphere cognisphere;
-
-  AgenticModel(Object object, Cognisphere cognisphere) {
-    super(object);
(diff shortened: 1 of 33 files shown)


experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModelFactory.java
```diff
@@ -45,57 +35,59 @@ public WorkflowModel fromAny(WorkflowModel prev, Object obj) {
 
   @Override
   public WorkflowModel combine(Map<String, WorkflowModel> workflowVariables) {
-    return new AgenticModel(workflowVariables, cognisphere);
+    throw new UnsupportedOperationException();
```
fjtirado: This means no parallel execution. Are we sure?

fluent/agentic-langchain4j/src/main/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowInvocationHandler.java
fjtirado, line 1: I thin this file should be moved  to experimental agentic, its not part of the fluent, but of the runtime

fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentTaskItemListBuilder.java
```diff
@@ -57,7 +57,8 @@ public AgentTaskItemListBuilder agent(String name, Object agent) {
         .forEach(
             exec ->
                 this.delegate.callFn(
-                    name, fn -> fn.function(AgentAdapters.toFunction(exec), Cognisphere.class)));
+                    name,
+                    fn -> fn.function(AgentAdapters.toFunction(exec), DefaultCognisphere.class)));
```
fjtirado: We really need this be the implementation object? What is in on DefaultCognisphere that is not in Cognisphere?

fluent/agentic-langchain4j/src/main/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowInvocationHandler.java
ricardozanini: We can't! Mainly because it has references to the `WorkflowBuilder`. In this case we will follow the orthogonal SPI from upstream. However, we could add a new module to have both DSL and Runtime parts. But that's too much added complexity for now to the architecture, IMO. We can review this later.

fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentTaskItemListBuilder.java
ricardozanini: SPI defines the `Object exec(DefaultCognisphere)`, so we must follow it.

experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModelFactory.java
ricardozanini: Fixed! (not properly, see my comment in the implementation)

experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModel.java
```diff
@@ -39,17 +35,15 @@ public void setObject(Object obj) {
 
   @Override
   public Collection<WorkflowModel> asCollection() {
-    return object instanceof Collection value
-        ? new AgenticModelCollection(value, cognisphere)
-        : Collections.emptyList();
+    throw new UnsupportedOperationException("Not supported yet.");
```
fjtirado, line 38: Ill fix this later after you merge

experimental/agentic/src/main/java/io/serverlessworkflow/impl/expressions/agentic/AgenticModel.java
```diff
@@ -39,17 +35,15 @@ public void setObject(Object obj) {
...
   @Override
   public <T> Optional<T> as(Class<T> clazz) {
-    if (Cognisphere.class.isAssignableFrom(clazz)) {
-      return Optional.of(clazz.cast(cognisphere));
-    } else if (ResultWithCognisphere.class.isAssignableFrom(clazz)) {
-      return Optional.of(clazz.cast(new ResultWithCognisphere<>(cognisphere, object)));
+    if (AgenticScope.class.isAssignableFrom(clazz)) {
+      return Optional.of(clazz.cast(object));
+    } else if (Map.class.isAssignableFrom(clazz)) {
+      return Optional.of(clazz.cast(((AgenticScope) object).state()));
```
fjtirado, line 51: hmmm, maybe we can do the same in asMap() method (overriding the one in Java model)

Issue #691, opened 2025-08-06
Add OAuth  authentication support to HTTP call task


treblereel:
HttpCallTask implementation lack OAuth authentication support

PR #718, opened 2025-08-19
Add OAuth authentication support to HTTP call task

treblereel:

https://github.com/serverlessworkflow/sdk-java/issues/691


api/src/test/java/io/serverlessworkflow/api/ApiTest.java
```diff
@@ -63,6 +63,23 @@ void testCallHTTPAPI() throws IOException {
...
+    Task task = workflow.getDo().get(0).getTask();
+    if (task.get() instanceof CallTask) {
+      CallTask callTask = task.getCallTask();
+      assertThat(callTask).isNotNull();
+      assertThat(task.getDoTask()).isNull();
+      CallHTTP httpCall = callTask.getCallHTTP();
+      assertThat(httpCall).isNotNull();
+      assertThat(httpCall.getWith().getMethod()).isEqualTo("get");
+    }
+  }
```
fjtirado: This one is already covered by https://github.com/serverlessworkflow/sdk-java/blob/main/api/src/test/java/io/serverlessworkflow/api/ApiTest.java#L84-L115, so can roolback changes on this file

impl/http/pom.xml
```diff
@@ -24,6 +24,10 @@
             <groupId>org.glassfish.jersey.core</groupId>
             <artifactId>jersey-client</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.fasterxml.jackson.core</groupId>
+            <artifactId>jackson-databind</artifactId>
+        </dependency>
```
fjtirado: Http module must NOT depend on jackson at compilation level.

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OAuth2AuthProvider.java
```diff
@@ -16,24 +16,53 @@
...
+    JWT token =
+        new OAuthRequestBuilder(workflowApplication, oauth2)
+            .build(workflow, task, model)
+            .validateAndGet();
+
+    String tokenType = (String) token.getClaim("typ");
+    builder.header(
+        AuthProviderFactory.AUTH_HEADER_NAME,
+        String.format(BEARER_TOKEN, tokenType, token.getToken()));
+  }
```
fjtirado: I think that OuathRequest should be built on build method and  preRequest should just call validateAndGet over that request and add the token to the header

impl/pom.xml
```diff
@@ -57,6 +58,11 @@
               <artifactId>jakarta.ws.rs-api</artifactId>
               <version>${version.jakarta.ws.rs}</version>
             </dependency>
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>${version.jackson-databind}</version>
+            </dependency>
```
fjtirado: The jackson version is already handled [here](https://github.com/serverlessworkflow/sdk-java/blob/main/pom.xml#L139-L142), not need to change this POM

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/HttpRequestBuilder.java
```diff
@@ -0,0 +1,127 @@
...
+    return this;
+  }
+
+  public HttpRequestBuilder withMethod(String method) {
+    this.method = method;
+    return this;
+  }
+
+  public HttpRequest build(WorkflowContext workflow, TaskContext task, WorkflowModel model) {
+    HttpRequest.Builder request = HttpRequest.newBuilder();
```
fjtirado: I see you are using java.net.http client to retrieve the token. However, HttpExecutor in the same module is using JaxRs client to invoke the endpoint
Since it does not make sense to have two different ways of calling a rest endpoint within the same module, I think this needs to be changed to use JaxRs.  
Once this is done, we can see how to avoid the compile dependency with Jackson to parse the token. We might need to do something similar to HttpModelConverter, but I think JaxRS client will gracefully handle the fact that jackson will be in the classpath. Lets see

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OAuth2AuthProvider.java
```diff
@@ -16,24 +16,53 @@
...
 import io.serverlessworkflow.impl.WorkflowModel;
+import io.serverlessworkflow.impl.executors.http.oauth.OAuthRequestBuilder;
+import jakarta.ws.rs.client.Invocation;
 import jakarta.ws.rs.client.Invocation.Builder;
 
 public class OAuth2AuthProvider implements AuthProvider {
 
+  private Oauth2 oauth2;
+
+  private final WorkflowApplication workflowApplication;
```
fjtirado: Rather than having these two private members, replace them by private OAuthRequestBuilder requestBuilder and initilize it passing the WorkflowApplication and OAuth2. 
Then, in prerequest you call build over this new member.

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/AccessTokenProvider.java
```diff
@@ -0,0 +1,67 @@
...
+    this.context = context;
+
+    ServiceLoader<JWTConverter> jwtConverters =
+        ServiceLoader.load(JWTConverter.class, AccessTokenProvider.class.getClassLoader());
+
+    if (jwtConverters.iterator().hasNext()) {
+      this.jwtConverter = jwtConverters.iterator().next();
+    } else {
+      throw new RuntimeException("No JWTConverter implementation found");
+    }
```
fjtirado: This can be a one line
```suggestion
        this.jwtConverter = ServiceLoader.load(JWTConverter.class).findFirst()
                .orElseThrow(() ->  new IllegalStateException("No JWTConverter implementation found"));
```

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/AccessTokenProvider.java
```diff
@@ -0,0 +1,67 @@
...
+  }
+
+  public JWT validateAndGet() {
+    Map<String, Object> token = tokenResponseHandler.apply(invocation, context);
+    JWT jwt;
+    try {
+      jwt = jwtConverter.fromToken((String) token.get("access_token"));
+    } catch (IllegalArgumentException e) {
+      throw new RuntimeException("Failed to parse JWT token: " + e.getMessage(), e);
+    }
```
fjtirado: Just let the IllegalArgumentException to escalate (converting one Runtime into another runtime is normally unnecesary)

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/AccessTokenProvider.java
```diff
@@ -0,0 +1,67 @@
...
+    if (!(issuers == null || issuers.isEmpty())) {
+      String tokenIssuer = (String) jwt.getClaim("iss");
+      if (tokenIssuer == null || tokenIssuer.isEmpty() || !issuers.contains(tokenIssuer)) {
+        throw new RuntimeException("Token issuer is not valid: " + tokenIssuer);
```
fjtirado: I think IllegalStateException is a better fit than generic RuntimeException here

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/HttpRequestBuilder.java
```diff
@@ -0,0 +1,148 @@
...
+  }
+
+  private void validate() {
+    if (uri == null) {
+      throw new IllegalStateException("URI must be set before building the request");
+    }
+
+    if (grantType == null) {
+      throw new IllegalStateException("Grant type must be set before building the request");
+    }
```
fjtirado: In case you do not know, when checking a single null pointer for throwing an exception with a message, you can  use
```suggestion
  Objects.requireNonNull(uri,"URI must be set before building the request");
  Objects.requireNonNull(grantType,"Grant type must be set before building the request");
```

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/InvocationHolder.java
```diff
@@ -0,0 +1,42 @@
...
+
+  Response invoke() {
+    return call.get();
+  }
+
+  void close() {
+    if (client != null) {
+      client.close();
+    }
+  }
```
fjtirado: Let use some standard java constructs here
```suggestion
class InvocationHolder implements Callable<Response>, Closeable{

  private final Client client;
  private final Supplier<Response> call;

  InvocationHolder(Client client, Supplier<Response> call) {
    this.client = client;
    this.call = call;
  }

  public Response call() {
    return call.get();
  }

  public void close() {
    if (client != null) {
      client.close();
    }
  }
```

impl/jwt-impl/pom.xml
```diff
@@ -0,0 +1,26 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.serverlessworkflow</groupId>
+        <artifactId>serverlessworkflow-impl</artifactId>
+        <version>8.0.0-SNAPSHOT</version>
+    </parent>
+
+    <artifactId>serverlessworkflow-impl-jwt</artifactId>
```
fjtirado: Lest reflect this is the jackson implementation of JWT in the artifact id
```suggestion
    <artifactId>serverlessworkflow-impl-jackson-jwt</artifactId>
```

impl/jwt-impl/src/main/java/io/serverlessworkflow/impl/http/jwt/DefaultJWTConverter.java
```diff
@@ -0,0 +1,45 @@
...
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import io.serverlessworkflow.http.jwt.JWT;
+import io.serverlessworkflow.http.jwt.JWTConverter;
+import java.nio.charset.StandardCharsets;
+import java.util.Base64;
+import java.util.Map;
+
+public class DefaultJWTConverter implements JWTConverter {
```
fjtirado:
```suggestion
public class JacksonJWTConverter implements JWTConverter {
```

impl/jwt-impl/src/main/java/io/serverlessworkflow/impl/http/jwt/DefaultJWTImpl.java
```diff
@@ -0,0 +1,44 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package io.serverlessworkflow.impl.http.jwt;
+
+import io.serverlessworkflow.http.jwt.JWT;
+import java.util.Map;
+
+public class DefaultJWTImpl implements JWT {
```
fjtirado:
```suggestion
public class JacksonJWTImpl implements JWT {
```

impl/pom.xml
```diff
@@ -14,6 +14,7 @@
         <version.net.thisptr>1.4.0</version.net.thisptr>
         <version.com.github.f4b6a3>5.2.3</version.com.github.f4b6a3>
         <version.jakarta.ws.rs>4.0.0</version.jakarta.ws.rs>
+        <version.jackson-databind>2.19.2</version.jackson-databind>
```
fjtirado: This line should not be needed, since jackson databind is managed in the parent pom

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/AccessTokenProvider.java
```diff
@@ -0,0 +1,62 @@
...
+  }
+
+  public JWT validateAndGet() {
+    Map<String, Object> token = tokenResponseHandler.apply(invocation, context);
+    JWT jwt;
+    try {
+      jwt = jwtConverter.fromToken((String) token.get("access_token"));
+    } catch (IllegalArgumentException e) {
+      throw new IllegalStateException("Failed to parse JWT token: " + e.getMessage(), e);
+    }
```
fjtirado: Why not letting the Illegal argument exception to escalate?. Unless Im missing something, adding an additional exception here, in my opinion, only pollutes the exception stack trace witn an additional one

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/TokenResponseHandler.java
```diff
@@ -0,0 +1,63 @@
...
+            WorkflowError.communication(
+                    response.getStatus(),
+                    context,
+                    "Failed to obtain token: HTTP "
+                        + response.getStatus()
+                        + " — "
+                        + response.getEntity())
+                .build());
+      }
+      return (Map<String, Object>) response.readEntity(Map.class);
```
fjtirado: You can easily avoid the casting
```suggestion
      return response.readEntity(new GenericType<Map<String,Object>>() {});
```

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/oauth/TokenResponseHandler.java
```diff
@@ -0,0 +1,63 @@
...
+    } catch (ResponseProcessingException e) {
+      throw new WorkflowException(
+          WorkflowError.communication(
+                  e.getResponse().getStatus(),
+                  context,
+                  "Failed to process response: " + e.getMessage())
```
fjtirado, line 50: Lets pass the exception here (I know that now its the same with current implementation, but in future we can take benefit of the stack trace information we are losing when passing the messa)
```suggestion
                 e
```

fjtirado, line 50: Lets pass the exception rather than the message (in future might be interesting to  have the stack trace
```suggestion
                 e)
```

jwt/src/main/java/io/serverlessworkflow/http/jwt/JWT.java
fjtirado, line 1: I was thinking about this interface. 
Since the token structure is a mix of standard (https://en.wikipedia.org/wiki/JSON_Web_Token#Standard_fields)  and custom fields, maybe we can leave the getClaim() for custom ones and add to the interface String typ() and String iss(), so we kill two birds with one shot, we do not have to call get("typ")   and we tell the implementos these are the fields we are using. wdyt?
Internally the jackson implementation can still be a map and return null when the key is not there.

fjtirado (approved):

PR #721, opened 2025-08-19
Use Spoteless for header checking and code-formatting

ricardozanini:

In this PR:

1. We replaced the old Spotify plugin and replaced header checks with Spotless.
2. Spotless now adds headers automatically to Java files
3. Added a Makefile to facilitate the local dev flow (run `make help`)
4. Added a pre-commit hook to format/add headers to new files in case you missed
5. Refactored the CONTRIB guidelines

@fjtirado if you agree, feel free to merge.

Since the project is growing, I believe we should start to improve the governance.

PS: I intend to change this formatter. The Google one is too narrow IMO.


ricardozanini:
I'll fix the error tomorrow. Actually, there's a bug on our CI. The `verify examples` is not running the samples. 😆

fjtirado:
@ricardozanini Besides that, I did a quick check in eclipse and Im not fully convinced about the makefile thing (user having to install the hook) I prefer Maven to do the whole job (this is dividing work and this usually not a good idea)
Also, after  doing make hooks,  I added a file without header in Eclipse, add to stage and the header was not added.  I still has to add it manually. 
Finally, and this is showstopper to me, mvn install at the root works, but mvn install at not root projects always fails with this error
Failed to execute goal com.diffplug.spotless:spotless-maven-plugin:2.43.0:check (default) on project serverlessworkflow-api: Execution default of goal com.diffplug.spotless:spotless-maven-plugin:2.43.0:check failed: Unable to locate file with path: /home/ftirados/git/sdk-java/api/config/license-header.txt: Could not find resource '/home/ftirados/git/sdk-java/api/config/license-header.txt'. -> [Help 1]
We do not want to execute mvn install for the whole project for any change.

PR #724, opened 2025-08-20
Add Email Drafter Human in the loop use case

ricardozanini:

In this PR:

1. Added the Email Drafter use case in the integration tests for Human in the loop pattern
2. Fix Event consumption on the DSL

Since this PR is already large and includes the use case and a fix to the implementation we identified while implementing it, I'll merge it as is and add documentation later.

### DIFF
--- a/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/ChatBotIT.java
+++ b/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/ChatBotIT.java
@@ -17,46 +17,41 @@ package io.serverlessworkflow.fluent.agentic;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.junit.jupiter.api.Assertions.assertEquals;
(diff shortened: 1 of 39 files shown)


fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/ChatBotIT.java
```diff
@@ -99,7 +89,9 @@ void chat_bot() {
                                                                 emit.event(
                                                                     e ->
                                                                         e.type(
-                                                                            "org.acme.chatbot.reply"))))))
+                                                                                "org.acme.chatbot.reply")
+                                                                            .data(
+                                                                                ".conversation"))))))
```
fjtirado: We will need also to pass a function to generate the cloud event,. EventData is part of EventProperties in Emit, so you basically already can pass a function there. You now need to change the fluent to pass a Function which returns a CloudEventData.

fjtirado (approved):

PR #804, opened 2025-09-23
Add OpenAPI task implementation

### LINKED ISSUE #477: Add OpenAPI task implementation
Implement [OpenAPI task](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#openapi-call). 

@mcruzdev Take a look to https://github.com/serverlessworkflow/sdk-java/blob/main/impl/http/src/main/java/io/serverlessworkflow/impl/executors/HttpExecutor.java
There should be a new maven module that follow the same approach for OpenAPI

treblereel:
[**Many thanks for submitting your Pull Request :heart:!**

Fix #477 

**Additional information (if needed):**]

### DIFF
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/ExpressionURISupplier.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/ExpressionURISupplier.java
@@ -15,8 +15,7 @@
  */
 package io.serverlessworkflow.impl.executors.http;
 
-import static io.serverlessworkflow.impl.executors.http.HttpExecutor.client;
-
+import io.serverlessworkflow.api.types.UriTemplate;
 import io.serverlessworkflow.impl.TaskContext;
 import io.serverlessworkflow.impl.WorkflowContext;
 import io.serverlessworkflow.impl.WorkflowModel;
@@ -32,6 +31,8 @@ public class ExpressionURISupplier implements TargetSupplier {
 
   @Override
   public WebTarget apply(WorkflowContext workflow, TaskContext task, WorkflowModel node) {
-    return client.target(expr.apply(workflow, task, node));
+    return HttpExecutor.getURISupplier(
+            new UriTemplate().withLiteralUriTemplate(expr.apply(workflow, task, node)))
+        .apply(workflow, task, node);
   }
 }
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java
@@ -47,7 +47,7 @@ import java.util.function.Supplier;
 
 public class HttpExecutor implements CallableTask<CallHTTP> {
 
-  public static final Client client = ClientBuilder.newClient();
+  private static final Client client = ClientBuilder.newClient();
 
   private TargetSupplier targetSupplier;
   private Optional<WorkflowValueResolver<Map<String, Object>>> headersMap;
--- a/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/HttpCallAdapter.java
+++ b/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/HttpCallAdapter.java
@@ -24,9 +24,6 @@ import io.serverlessworkflow.api.types.HTTPQuery;
 import io.serverlessworkflow.api.types.Headers;
 import io.serverlessworkflow.api.types.Query;
 import io.serverlessworkflow.api.types.ReferenceableAuthenticationPolicy;
-import io.serverlessworkflow.api.types.TaskTimeout;
-import io.serverlessworkflow.api.types.Timeout;
-import io.serverlessworkflow.api.types.TimeoutAfter;
 import io.serverlessworkflow.api.types.UriTemplate;
 import io.swagger.v3.oas.models.media.Schema;
 import io.swagger.v3.oas.models.parameters.Parameter;
@@ -84,14 +81,6 @@ class HttpCallAdapter {
     addBody(httpArgs);
 
     addTarget(endpoint);
-
-    TaskTimeout taskTimeout = new TaskTimeout();
-    Timeout timeout = new Timeout();
-    taskTimeout.withTaskTimeoutDefinition(timeout);
-    TimeoutAfter timeoutAfter = new TimeoutAfter();
-    timeout.setAfter(timeoutAfter);
-    timeoutAfter.withDurationExpression("PT30S");
-    callHTTP.setTimeout(taskTimeout);
 
     return callHTTP;
   }


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/ExpressionURISupplier.java
fjtirado, line 1: Since this class is already defined here https://github.com/serverlessworkflow/sdk-java/blob/main/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java#L207-L218
I think we should make it public and reuse here

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
```diff
@@ -0,0 +1,154 @@
...
+    String operationId = task.getWith().getOperationId();
+    URI openAPIEndpoint = targetSupplier.apply(workflowContext, taskContext, input);
+    OpenAPIProcessor processor = new OpenAPIProcessor(operationId, openAPIEndpoint);
+    OperationDefinition operation = processor.parse();
+
+    OperationPathResolver pathResolver =
+        new OperationPathResolver(
+            operation.getPath(),
+            application,
+            task.getWith().getParameters().getAdditionalProperties());
```
fjtirado: I think all this stuff should be moved to init method (thy are always the same for different executions, so they only need to be processed once)

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
```diff
@@ -0,0 +1,154 @@
...
+                  .redirect(task.getWith().isRedirect())
+                  .target(pathResolver.resolve(workflowContext, taskContext, input))
+                  .workflowParams(task.getWith().getParameters().getAdditionalProperties());
+
+          WorkflowException workflowException = null;
+
+          for (var server : operation.getServers()) {
+            CallHTTP callHTTP = httpCallAdapter.server(server).build();
+            HttpExecutor executor = new HttpExecutor();
+            executor.init(callHTTP, workflow, application, resourceLoader);
```
fjtirado: Same hre the HttpExecutor,init should be called on OpenApiExecutor.init. 
The idea is to keep the apply method as minimum as possible

treblereel: I’ve significantly improved the class, but unfortunately I can’t do this in all cases because I need (WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input), which are only available in the apply() method

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/HttpCallAdapter.java
```diff
@@ -0,0 +1,211 @@
...
+    addBody(httpArgs);
+
+    addTarget(endpoint);
+
+    TaskTimeout taskTimeout = new TaskTimeout();
+    Timeout timeout = new Timeout();
+    taskTimeout.withTaskTimeoutDefinition(timeout);
+    TimeoutAfter timeoutAfter = new TimeoutAfter();
+    timeout.setAfter(timeoutAfter);
+    timeoutAfter.withDurationExpression("PT30S");
```
fjtirado: This should be configurable (or deleted)

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
fjtirado: I do not understand why you cannot process the schema file in the init method if the endpoint uri of the schema is static.

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
```diff
@@ -0,0 +1,128 @@
...
+  public boolean accept(Class<? extends TaskBase> clazz) {
+    return clazz.equals(CallOpenAPI.class);
+  }
+
+  @Override
+  public CompletableFuture<WorkflowModel> apply(
+      WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input) {
+
+    OperationDefinitionSupplier operationDefinitionSupplier =
+        new OperationDefinitionSupplier(application, task);
```
fjtirado: OperationDefinitionSupplier should be moved to init. 
And I think the parsing of the openapi file can be done in OperationDefinitionSupplier, (If I interpreted the specification properly, the runtime of the schema is an static resource, not a dynamic one) so you do the parsing of the schema file only once, not for every call
The assumption Im doing here is that the URI of the openapi schema is fixed https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#external-resource
The tricky part is authorization (which refers to authorization to access the schema). Authroization might be based on expressions (and expression evaluation requires WorkflowContext, TaskContext and the model) but I think in that case we should assume that the authorization part is also fixed. 
Im pretty reluctant to parse the openapi file every time the function is called. So, if you prefer to reuse the existing auth code (which certainly and prorperly requires WorkflowContect and TaskContext) then you should cache the result of the parsing so subsequent calls does not retrieve the file again. But (since at the end, when you cache to only access the schema one, you will be assuming that the auth info remains constant)  I think is worth to explore a refactor of the auth code in a way that it can be executed  assuming there are not expressions in the auth strings.

fjtirado: Or we can say that we accessing the schema every time the function is called in case the schema has changed, but this is dangerous because the reality is that the user invoking the schema has defined the workflow (the input and output of the call) based on a particular schema version. It is kind of weird that the schema changes after writing the flow, it will be like compiling java program against a dynamically changing java library.  I have raised the point in the specs forums

fjtirado: So, for now, just move OperationDefinitionSupplier to the init, plus anything else that you detect does not depend on the runtime context (aka: does not require  expression evaluation). We can always refactor later assuming a static schema URI.

fjtirado: After checking in the CNCF forums, my assumption that the endpoint is static is wrong. 
The endpoint indeed might change. 
What I feel we should do then, is to use HTTP cache through HEAD to avoid parsing the same file again and again if the endpoint does not change. 
We can do that with a different issue, but I want to be sure we are aligned on this

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java
```diff
@@ -47,7 +47,7 @@
 
 public class HttpExecutor implements CallableTask<CallHTTP> {
 
-  private static final Client client = ClientBuilder.newClient();
+  public static final Client client = ClientBuilder.newClient();
```
fjtirado: Does client variable needs to be public?
I did not find any place where it is used.

treblereel: it was ExpressionURISupplier, i improved that

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
```diff
@@ -0,0 +1,125 @@
...
+            HttpExecutor executor = new HttpExecutor();
+            executor.init(callHTTP, definition);
+
+            try {
+              return executor.apply(workflowContext, taskContext, input).get();
+            } catch (WorkflowException e) {
+              workflowException = e;
+            } catch (Exception e) {
+              throw new RuntimeException(e);
+            }
```
fjtirado, line 77: You do not need this block, you just want to propagate all exception that are not workflow (because if it workflow you want to try the next server), which will happen anyway if you delete this block

treblereel: executor.apply returns a CompletableFuture, and get() declares InterruptedException and ExecutionException.

I don’t see anything wrong with wrapping these exceptions in a RuntimeException and throwing it.

mcruzdev:
Hi @treblereel, nice pull request!

I would like to mention the redirect option, are you planning to do in another pull request?

mcruzdev:
To be more specific, this one https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#properties-6 _redirect_

treblereel:
@mcruzdev OpenAPI is quite a large topic, and I prefer to work on it in small iterations. It makes reviewing and testing easier

ricardozanini:
@mcruzdev feel free to add it - I'm merging this one.

mcruzdev:
> @mcruzdev OpenAPI is quite a large topic, and I prefer to work on it in small iterations. It makes reviewing and testing easier

I agree with you.

> @mcruzdev feel free to add it - I'm merging this one.

Yeah, I can work on it.

fjtirado:
@mcruzdev @treblereel Ill be heavily refactoring openAPI implementation those days because of https://github.com/serverlessworkflow/sdk-java/issues/879, so its better to wait to add more functionality for the time being. 
@mcruzdev in the meantime you can start looking to https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#shell-process, which is a long and independent task

PR #806, opened 2025-09-23
[Fix #805] WorkflowDefinitionId should be public

### LINKED ISSUE #805: WorkflowId should be public
It might be interesting to access a workflow definition given its WorkflowId.

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/805

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
@@ -16,7 +16,6 @@
 package io.serverlessworkflow.impl;
 
 import com.github.f4b6a3.ulid.UlidCreator;
-import io.serverlessworkflow.api.types.Document;
 import io.serverlessworkflow.api.types.SchemaInline;
 import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.impl.events.EventConsumer;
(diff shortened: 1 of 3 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowInstanceIdFactory.java
fjtirado, line 1: Renamed to make explicit that this factory is not related with the definition id, but with the instance id

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -249,15 +248,13 @@ public WorkflowApplication build() {
     }
   }
 
-  private static record WorkflowId(String namespace, String name, String version) {
-    static WorkflowId of(Document document) {
-      return new WorkflowId(document.getNamespace(), document.getName(), document.getVersion());
-    }
+  public Map<WorkflowDefinitionId, WorkflowDefinition> workflowDefinitions() {
```
fjtirado, line 251: Application can now be used as a workflow definition registry

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowDefinitionId.java
```diff
@@ -0,0 +1,35 @@
...
+  public static final String DEFAULT_VERSION = "0.0.1";
+  public static final String DEFAULT_NAMESPACE = "org.acme";
+
+  public static WorkflowDefinitionId of(Workflow workflow) {
+    Document document = workflow.getDocument();
+    return new WorkflowDefinitionId(
+        document.getNamespace(), document.getName(), document.getVersion());
+  }
+
+  public static WorkflowDefinitionId fromName(String name) {
```
fjtirado, line 32: Useful as syntactic sugar in case you want to identify workflows with a single name. 
If not use the record constructor to specify namespace and version

Issue #732, opened 2025-08-25
Add Ollama to the optional integration test pipeline on PR checks to run the agentic use cases


ricardozanini:
**What would you like to be added**:
Our `*IT` tests are not running on PR checks since it may slow down or cycle due to the time needed to setup and run Ollama on a small container such as GHA.

This PR is already setup, it can run on demand by the maintainers by commenting on the PR `run /integration-tests`

**Why is this needed**:
Improve the overall code reliability.

PR #808, opened 2025-09-23
Agentic workflow examples

treblereel:
https://github.com/serverlessworkflow/sdk-java/issues/809


experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java
fjtirado, line 1: I think this one should be disable by default ( I mean, not run every time we do mvn clean install over the SDK)
Im not sure thats the case, hence this comment
@ricardozanini

ricardozanini: It's not. All the IT tests only run when we do `-Pintegration-tests`. See #732

experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java
```diff
@@ -0,0 +1,288 @@
...
+    var a2 = AgentsUtils.newTechnicalExpert();
+    var a3 = AgentsUtils.newLegalExpert();
+
+    Workflow wf =
+        workflow("conditional")
+            .sequence("process", category)
+            .tasks(
+                t ->
+                    t.switchCase(
+                        p ->
```
ricardozanini: We don't need this branch. On `UNKNOWN` the workflow just won't call any agent. What you can do is a `when(UNKNOWN).emit(eventAlertUnknown)`.

experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java
```diff
@@ -0,0 +1,288 @@
...
+    Workflow wf =
+        AgentWorkflowBuilder.workflow("maxFlow")
+            .tasks(
+                d ->
+                    d.loop(
+                        "limit",
+                        l ->
+                            l.maxIterations(5)
+                                .exitCondition(c -> c.readState("score", 0).doubleValue() >= 0.8)
+                                .subAgents("sub", scorer, editor)))
```
ricardozanini: Here we should use the shortcuts via `DSL.loop`. You may send another parallel PR to include `maxIterations` to the shortcuts API if needed.

experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java
```diff
@@ -0,0 +1,288 @@
...
+public class LC4JEquivalenceIT {
+
+  @Test
+  @DisplayName("Sequential agents via DSL.sequence(...)")
+  public void sequentialWorkflow() {
+    var a1 = AgentsUtils.newCreativeWriter();
+    var a2 = AgentsUtils.newAudienceEditor();
+    var a3 = AgentsUtils.newStyleEditor();
+
+    Workflow wf = workflow("seqFlow").tasks(tasks -> tasks.sequence("process", a1, a2, a3)).build();
```
ricardozanini: instead of `a1, a2, a3`, let's give the agents a name so users can grasp it better.

experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/README.md
```diff
@@ -0,0 +1,10 @@
...
+| Conditional workflow  | [Conditional workflow](https://docs.langchain4j.dev/tutorials/agents/#conditional-workflow) | [LC4JEquivalenceIT.java#L218](https://github.com/treblereel/serverless-workflow-sdk-java/blob/langchain_examples/experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java#L218)                                           |
+| Error handling        | [Error handling](https://docs.langchain4j.dev/tutorials/agents/#error-handling)           | [LC4JEquivalenceIT.java#L185](https://github.com/treblereel/serverless-workflow-sdk-java/blob/langchain_examples/experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java#L185)                                           |
+| Human-in-the-loop     | [Human-in-the-loop](https://docs.langchain4j.dev/tutorials/agents/#human-in-the-loop)     | [LC4JEquivalenceIT.java#L257](https://github.com/treblereel/serverless-workflow-sdk-java/blob/langchain_examples/experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java#L257)                                           |
```
ricardozanini: Instead of this table, can you add the code excerpt? So the reader doesn't need to click on each link and go back and forth.

ricardozanini (commented):
Thanks! A few comments if you don't mind!

experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/LC4JEquivalenceIT.java
ricardozanini: We can even improve the shortcuts:
```
loop("reviewLoop").until(conditional).agents(scorer, editor).
```

wdyt?

PR #815, opened 2025-09-25
[Fix #803] Fixing suspend/resume with parallel

### LINKED ISSUE #803: Suspend/resume wont work properly with ForkTask
Current implementation is only valid for one suspended task at a given time, which might not be the case when using ForkTask

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/803


impl/test/src/test/java/io/serverlessworkflow/impl/test/ForkWaitTest.java
```diff
@@ -0,0 +1,82 @@
...
+  void testForkWait() throws IOException, InterruptedException, ExecutionException {
+    assertModel(
+        appl.workflowDefinition(readWorkflowFromClasspath("workflows-samples/fork-wait.yaml"))
+            .instance(Map.of())
+            .start()
+            .join());
+  }
+
+  @Test
+  void testForkWaitWithSuspend() throws IOException, InterruptedException {
```
fjtirado, line 60: This test was failing without the fix

impl/test/src/test/java/io/serverlessworkflow/impl/test/ForkWaitTest.java
```diff
@@ -0,0 +1,82 @@
...
+    CompletableFuture<WorkflowModel> future = instance.start();
+    Thread.sleep(50);
+    assertThat(instance.status()).isEqualTo(WorkflowStatus.WAITING);
+    instance.suspend();
+    assertThat(instance.status()).isEqualTo(WorkflowStatus.SUSPENDED);
+    Thread.sleep(200);
+    instance.resume();
+    WorkflowModel model = future.join();
+    assertThat(instance.status()).isEqualTo(WorkflowStatus.COMPLETED);
+    assertModel(model);
```
ricardozanini, line 72: I'd use awaitability here next time we make a change. :p

ricardozanini (approved):
Comment's not a blocker.

PR #816, opened 2025-09-25
Upgrade LangChain4j 1.5.0; Fix deprecated warnings

ricardozanini:

See #792 

- Fixed a few deprecation warnings
- Fixed assertions imports


impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowModelFactory.java
```diff
@@ -70,10 +71,6 @@ default WorkflowModel fromAny(Object obj) {
       return from((Map<String, Object>) obj);
     } else if (obj instanceof WorkflowModel model) {
       return model;
-    } else if (obj instanceof CloudEventData ce) {
-      return from(ce);
-    } else if (obj instanceof CloudEvent ce) {
-      return from(ce);
```
fjtirado, line 76: Ill keep these two else-if just in case

fjtirado: ok duplicated

PR #817, opened 2025-09-25
Use ulid library safe for graalvm native compilation

### LINKED ISSUE #812: Native-image build fails due to UlidCreator static initialization (Random in image heap)
## Summary
Building a native executable with GraalVM/Mandrel fails because the SDK’s default ID generation uses `com.github.f4b6a3.ulid.UlidCreator`, which initializes a `SecureRandom` during class initialization. This creates a `Random` instance in the image heap, and native-image aborts.

## Current behavior

Default ID generation is defined in `WorkflowApplication.Builder`:

```java
// current
private WorkflowInstanceIdFactory idFactory =
    () -> com.github.f4b6a3.ulid.UlidCreator.getMonotonicUlid().toString();
```

During native image analysis, `UlidCreator$MonotonicFactoryHolder` runs a static initializer that constructs a `SecureRandom`, which is then detected in the image heap.

**Typical error (excerpt):**

```
Error: Detected an instance of Random/SplittableRandom class in the image heap.
The culprit object has been instantiated by the 'com.github.f4b6a3.ulid.UlidCreator$MonotonicFactoryHolder' class initializer...
    at java.security.SecureRandom.<init>(SecureRandom.java:224)
    at com.github.f4b6a3.ulid.UlidFactory$ByteRandom.newRandomFunction(UlidFactory.java:460)
    ...
    at com.github.f4b6a3.ulid.UlidCreator$MonotonicFactoryHolder.<clinit>(UlidCreator.java:165)
```

## Expected behavior

Native builds should succeed out of the box; any RNG instantiation should happen at **runtime**, not during image build.

---

## Root cause

`UlidCreator.getMonotonicUlid()` relies on a static holder which initializes a `SecureRandom` in a class initializer. GraalVM/Mandrel forbids `Random/SecureRandom` instances in the image heap, hence the failure.

---

### **Option 1 (keep ULID): lazy-initialize at runtime**

Avoid `UlidCreator` (static singleton). Instead, construct an `UlidFactory` instance at **runtime** (e.g., inside `build()`), and use it to generate IDs. Do **not** store a static factory or RNG.

**Suggested change:**

```diff
- private WorkflowInstanceIdFactory idFactory =
-     () -> com.github.f4b6a3.ulid.UlidCreator.getMonotonicUlid().toString();
+ private WorkflowInstanceIdFactory idFactory; // set lazily in build()

 public WorkflowApplication build() {
   ...
+  if (idFactory == null) {
+    // Build the ULID factory at runtime (no static holder)
+    final com.github.f4b6a3.ulid.UlidFactory f =
+        com.github.f4b6a3.ulid.UlidFactory.newMonotonicInstance();
+    idFactory = () -> f.create().toString();
+  }
   return new WorkflowApplication(this);
 }
```

**Pros**

* Preserves ULID format (monotonic, lexicographically sortable).
* Minimal behavioral change; no downstream visible differences besides native compatibility.

**Cons**

* Still depends on `ulid-creator` (which is fine if we’re happy with it).

---

### **Option 2 (replace with UUID): use JDK UUIDs by default**

Remove the `ulid-creator` dependency and use `UUID.randomUUID()` as the default. Keep `withIdFactory(...)` so advanced users can plug their own ULID (or anything else) if needed.

**Suggested change:**

```diff
- import com.github.f4b6a3.ulid.UlidCreator;
+ import java.util.UUID;

- private WorkflowInstanceIdFactory idFactory =
-     () -> UlidCreator.getMonotonicUlid().toString();
+ private WorkflowInstanceIdFactory idFactory; // set lazily in build()

 public WorkflowApplication build() {
   ...
+  if (idFactory == null) {
+    idFactory = () -> UUID.randomUUID().toString();
+  }
   return new WorkflowApplication(this);
}
```

**Pros**

* No extra dependency; JDK-only and Graal-friendly.
* Simplest path; avoids ULID static init entirely.

**Cons**

* ID format changes (not lexicographically sortable).
* Downstream code or docs expecting ULIDs would need to adjust (if any).

---

## Backward compatibility / migration notes

* The SDK already exposes `withIdFactory(WorkflowInstanceIdFactory)` on the builder, so advanced users can keep ULIDs (or any other scheme).
* If keeping ULID is important, Option 1 preserves the format.
* If switching to UUID (Option 2), call out ID format change in release notes.

---

## Acceptance criteria

* Native builds complete successfully without additional flags (`--initialize-at-run-time=...`).
* No `Random/SecureRandom` instances are created during image build (only at runtime).
* Existing public API remains intact; existing apps continue to run unchanged (besides the ID format if Option 2 is chosen).

---

## Tests & verification

* Add/adjust a smoke test that starts the engine and creates at least one workflow instance in a native test environment.
* (Optional) Add a unit test guarding against class initialization side-effects: ensure no references to `UlidCreator` are needed unless a custom ULID `IdFactory` is explicitly provided.

---

## Proposed default

Either option works; **Option 1** keeps existing semantics (ULID) with minimal change, while **Option 2** simplifies dependencies. The team can choose based on preference for ULID properties vs. dependency reduction.

mcruzdev:

* Remove the  `com.github.f4b6a3:ulid-creator` dependency.
* Add a new dependency `de.huxhorn.sulky:de.huxhorn.sulky.ulid` for ULID generation.
* Add a default `WorkflowInstanceIdFactory` implementation (`UlidWorkflowInstanceIdFactory`)

Closes #812

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
@@ -15,7 +15,6 @@
  */
 package io.serverlessworkflow.impl;
 
-import com.github.f4b6a3.ulid.UlidCreator;
 import io.serverlessworkflow.api.types.SchemaInline;
 import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.impl.events.EventConsumer;
@@ -31,6 +30,7 @@ import io.serverlessworkflow.impl.resources.ResourceLoaderFactory;
 import io.serverlessworkflow.impl.resources.StaticResource;
 import io.serverlessworkflow.impl.schema.SchemaValidator;
 import io.serverlessworkflow.impl.schema.SchemaValidatorFactory;
+import java.security.SecureRandom;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
@@ -141,7 +141,8 @@ public class WorkflowApplication implements AutoCloseable {
     private ResourceLoaderFactory resourceLoaderFactory = DefaultResourceLoaderFactory.get();
     private SchemaValidatorFactory schemaValidatorFactory;
     private WorkflowPositionFactory positionFactory = () -> new QueueWorkflowPosition();
-    private WorkflowInstanceIdFactory idFactory = () -> UlidCreator.getMonotonicUlid().toString();
+    private final SecureRandom secureRandom = new SecureRandom();
+    private WorkflowInstanceIdFactory idFactory = new UlidWorkflowInstanceIdFactory();
     private ExecutorServiceFactory executorFactory = new DefaultExecutorServiceFactory();
     private EventConsumer<?, ?> eventConsumer;
     private Collection<EventPublisher> eventPublishers = new ArrayList<>();
--- a/impl/core/pom.xml
+++ b/impl/core/pom.xml
@@ -21,8 +21,8 @@
             <artifactId>slf4j-api</artifactId>
         </dependency>
         <dependency>
-            <groupId>com.github.f4b6a3</groupId>
-            <artifactId>ulid-creator</artifactId>
+            <groupId>de.huxhorn.sulky</groupId>
+            <artifactId>de.huxhorn.sulky.ulid</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/UlidWorkflowInstanceIdFactory.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/UlidWorkflowInstanceIdFactory.java
@@ -1,0 +1,33 @@
+/*
+ * Copyright 2020-Present The Serverless Workflow Specification Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.impl;
+
+import de.huxhorn.sulky.ulid.ULID;
+import java.security.SecureRandom;
+
+/**
+ * A {@link WorkflowInstanceIdFactory} implementation that generates ULIDs as workflow instance IDs.
+ */
+public class UlidWorkflowInstanceIdFactory implements WorkflowInstanceIdFactory {
+
+  private final SecureRandom random = new SecureRandom();
+  private final ULID ulid = new ULID(random);
+
+  @Override
+  public String get() {
+    return ulid.nextULID();
+  }
+}
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -9,10 +9,10 @@
     <name>Serverless Workflow :: Impl</name>
     <packaging>pom</packaging>
     <properties>
-        <version.org.glassfish.jersey>3.1.11</version.org.glassfish.jersey>
-        <version.net.thisptr>1.6.0</version.net.thisptr>
-        <version.com.github.f4b6a3>5.2.3</version.com.github.f4b6a3>
+        <version.de.huxhorn.sulky>8.3.0</version.de.huxhorn.sulky>
         <version.jakarta.ws.rs>4.0.0</version.jakarta.ws.rs>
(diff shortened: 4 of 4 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -141,7 +140,7 @@ public SchemaValidator getValidator(SchemaInline inline) {
     private ResourceLoaderFactory resourceLoaderFactory = DefaultResourceLoaderFactory.get();
     private SchemaValidatorFactory schemaValidatorFactory;
     private WorkflowPositionFactory positionFactory = () -> new QueueWorkflowPosition();
-    private WorkflowInstanceIdFactory idFactory = () -> UlidCreator.getMonotonicUlid().toString();
+    private WorkflowInstanceIdFactory idFactory = new MonotnicUlidWorkflowInstanceIdFactory();
```
ricardozanini: I think we can keep using the Supplier, no?

mathecruz: The reason I created this class is that every time a ULID is instantiated, it also creates a new SecureRandom, which is expensive.

impl/core/src/main/java/io/serverlessworkflow/impl/MonotonicUlidWorkflowInstanceIdFactory.java
```diff
@@ -0,0 +1,40 @@
...
+
+  @Override
+  public synchronized String get() {
+    if (previousUlid == null) {
+      previousUlid = ulid.nextValue();
+    } else {
+      previousUlid = ulid.nextMonotonicValue(previousUlid);
+    }
+    return previousUlid.toString();
+  }
```
fjtirado: I think this should be more lazy. 
Reason is that, with this implementaiton, if the user is replacing this factory, the secure ramdon and the ULID are still created, but never used. 
There are two possibilities to achieve that:
- creare this class, as it is, (well, secureRamdon does not need to be a member in any case) in a lazy way in the WorkflowApplication (so create it only when the user has not set another WorkflowInsanceIdFactory)
- Leave WorkflowApplication as it is and create the ULID (and the SecureRamdon) when previousUlid is null.

fjtirado: I now explain why I prefer the first one (lazy initialization of the factory in WorkflowApplication)

The get method can be simplified by creating previousUlid in the constructor, so you write a not synchronized get. 

```
@Override
// note that you do not need to syncronized this
  public String get() {
       String result=  currentUlid.toString();
      currentUlid = ulid.nextMonotonicValue(currentUlid);
      return result;
  }

```
The constructor will look like

```
private final ULID ulid;
private ULID.Value currentUlid;

public MonotonicUlidWorkflowInstanceIdFactory () {
   ulid = new ULID(new SecureRandom());
   currentUlid = ulid.nextValue();
}

```
And in workflowapplicationbuilder.build you set the factory to MonoloictUlidWorkflowInstanceFactory if the id factory it is null (the user has not set its own one) rather than in builder initialization

fjtirado (commented):
@mcruzdev Posible solution proposed, please take a look

mcruzdev:
> Reason is that, with this implementaiton, if the user is replacing this factory, the secure ramdon and the ULID are still created, but never used.

Good catch!

> The get method can be simplified by creating previousUlid in the constructor, so you write a not synchronized get.

@fjtirado, do we need to guarantee no duplicated keys in a thread-safe way? If not, I agree with you to remove the `synchronized` or similar locks.

fjtirado:
@mcruzdev 
To avoid synchronization (and the risk derived from not using it), we can write

```
private final AtomicReference<ULID.Value> currentUlid;

public MonotonicUlidWorkflowInstanceIdFactory () {
   ulid = new ULID(new SecureRandom());
   currentUlid = new AtomicReference<>(ulid.nextValue());
}
 public String get() {   
      return currentUlid.getAndSet(ulid.nextMonotonicValue(currentUlid));
  }

```

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -141,7 +140,7 @@ public SchemaValidator getValidator(SchemaInline inline) {
     private ResourceLoaderFactory resourceLoaderFactory = DefaultResourceLoaderFactory.get();
     private SchemaValidatorFactory schemaValidatorFactory;
     private WorkflowPositionFactory positionFactory = () -> new QueueWorkflowPosition();
-    private WorkflowInstanceIdFactory idFactory = () -> UlidCreator.getMonotonicUlid().toString();
+    private WorkflowInstanceIdFactory idFactory = new MonotonicUlidWorkflowInstanceIdFactory();
```
fjtirado: @mcruzdev Remember that, in order to prevent SecureReandom instance to be created if the default id factory is overriden, this line (creation of the monotonic instance) neeeds to be moved to build() method

PR #835, opened 2025-09-29
Added AgenticServices helper

treblereel:


experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
```diff
@@ -37,38 +52,38 @@ void sequential_agents_tests() {
     WorkflowAgentsBuilder builder = new LC4JWorkflowBuilder();
 
     CreativeWriter creativeWriter =
-        spy(
-            AgenticServices.agentBuilder(CreativeWriter.class)
-                .chatModel(BASE_MODEL)
-                .outputName("story")
-                .build());
+            spy(
+                    dev.langchain4j.agentic.AgenticServices.agentBuilder(CreativeWriter.class)
```
ricardozanini: We can drop the FQCN

experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgenticServices.java
```diff
@@ -0,0 +1,125 @@
...
+        throw new IllegalArgumentException(runner + " must be an interface to create a Proxy");
+      }
+
+      ClassLoader cl = runner.getClassLoader();
+      Class<?>[] ifaces = new Class<?>[] {runner};
+      return (T) Proxy.newProxyInstance(cl, ifaces, h);
+    }
+  }
+
+  private static class AgentInvocationHandler implements InvocationHandler {
```
ricardozanini: I think we already have something pretty similar on the `agentic-langchain4j` module. Is it doable to share the same class?

experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaFuncUtils.java
```diff
@@ -32,6 +33,9 @@ static Object predObject(Predicate<?> pred, Optional<Class<?>> predClass) {
   }
 
   static <T> T convertT(WorkflowModel model, Optional<Class<T>> inputClass) {
+    if (model instanceof HasAgenticScope<?> hasAgenticScope && inputClass.isEmpty()) {
+      return (T) hasAgenticScope.getAgenticScope();
+    }
```
fjtirado: You should not do this change here, but in `as` (when inputclass is not empty) and` asJavaObject` (when inputclass is empty)  methods of AgenticModel

experimental/lambda/src/main/java/io/serverlessworkflow/impl/expressions/func/HasAgenticScope.java
fjtirado, line 1: This class should not be added, the change in convert should be done in asJavaObject

experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
```diff
@@ -38,21 +56,21 @@ void sequential_agents_tests() {
 
     CreativeWriter creativeWriter =
         spy(
-            AgenticServices.agentBuilder(CreativeWriter.class)
+            dev.langchain4j.agentic.AgenticServices.agentBuilder(CreativeWriter.class)
```
ricardozanini: We should remove the FQCN

experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
```diff
@@ -77,4 +95,123 @@ void sequential_agents_tests() {
...
   }
+
+  @Test
+  public void sequenceHelperTest() {
+    var creativeWriter = newCreativeWriter();
+    var audienceEditor = newAudienceEditor();
+    var styleEditor = newStyleEditor();
+
+    AgentsUtils.NovelCreator novelCreator =
+        io.serverlessworkflow.fluent.agentic.AgenticServices.of(AgentsUtils.NovelCreator.class)
```
ricardozanini: There are a few places with FQCN.

experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
```diff
@@ -77,4 +95,123 @@ void sequential_agents_tests() {
...
+  @Test
+  public void agentAndSequenceHelperTest() {
+    var creativeWriter = newCreativeWriter();
+    var audienceEditor = newAudienceEditor();
+    var styleEditor = newStyleEditor();
+
+    AgentsUtils.NovelCreator novelCreator =
+        io.serverlessworkflow.fluent.agentic.AgenticServices.of(AgentsUtils.NovelCreator.class)
+            .flow(workflow("seqFlow").agent(creativeWriter).sequence(audienceEditor, styleEditor))
+            .build();
```
ricardozanini, line 124: This DSL is looking pretty cool. We should do the same for `FuncWorkflowBuilder`, which is the one we are using on `quarkus-flow` next.

experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
```diff
@@ -77,4 +95,123 @@ void sequential_agents_tests() {
...
+            return holder;
+          }
+        };
+
+    String result =
+        AgenticServices.of(Agents.HoroscopeAgent.class)
+            .flow(
+                workflow("humanInTheLoop")
+                    .inputFrom(askSign)
+                    // .tasks(tasks -> tasks.callFn(fn(askSign))) // TODO should work too
```
ricardozanini, line 211: Can you open an issue to track this?

experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgenticWorkflow.java
```diff
@@ -0,0 +1,134 @@
...
+        throw new IllegalArgumentException(runner + " must be an interface to create a Proxy");
+      }
+
+      ClassLoader cl = runner.getClassLoader();
+      Class<?>[] ifaces = new Class<?>[] {runner};
+      return (T) Proxy.newProxyInstance(cl, ifaces, h);
+    }
+  }
+
+  private class AgentInvocationHandler implements InvocationHandler {
```
ricardozanini, line 72: This class is similar to this one: https://github.com/serverlessworkflow/sdk-java/blob/main/experimental/fluent/agentic-langchain4j/src/main/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowInvocationHandler.java#L37

We should bring them together.

ricardozanini (commented):
It's getting pretty neat!

experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
treblereel: @ricardozanini The issue is that AgenticServices conflicts with AgenticServices from langchain4g. Maybe we can come up with a different name instead of using the one Mark suggested? 

For example, simply AgenticServicesFlow?

experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgenticWorkflow.java
treblereel: In my personal opinion, they aren’t equivalent. But if the task is to use the InvocationHandler from WorkflowInvocationHandler.java here, I can give it a try

experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
ricardozanini: I think `AgenticWorkflow` can be a better option, wdyt?

experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgenticWorkflow.java
ricardozanini: They are not equivalent, but they are doing similar things that we can share between them, wdyt?

experimental/fluent/agentic-langchain4j/src/test/java/io/serverlessworkflow/fluent/agentic/langchain4j/WorkflowAgentsIT.java
treblereel: Or just AgenticFlow ?

Issue #836, opened 2025-09-30
Implement `beforeAgentInvocation` and `afterAgentInvocation` for LC4J 1.6.0


ricardozanini:
**What would you like to be added**:
After upgrading to LC4J 1.6.0 our implementation of their workflow API is missing the handlers for before and after agent invocation.

This can be mapped to our `inputFrom` (before) and `outputAs` after for each call agent task.

There are new objects introduced by the API named `AgentRequest` and `AgentReponse` that carry the input and output, respectively, to the calls that we must handle with `AgenticScope`.

**Why is this needed**:
To be 100% compatible with their API.

As a workaround, one can use our DSL instead and have more control of each agent call by implementing their own `Function` on `inputFrom` and `outputAs`. For example:

```java
AgentWorkflowBuilder.workflow().tasks(t -> t.agent(agent).inputFrom(AgenticScope -> {}).outputAs(AgenticScope -> {}))
```

And infer directly into the input/output of the agent task. Simple.

matheusandre1:
@ricardozanini Can I do that?

ricardozanini:
Oh this one is outdated, we are not using lc4j any longer in this project. Need to close this!

PR #837, opened 2025-09-30
Adapt quarkus

ricardozanini:
Manually upgrade due to API breaking, see #834


ricardozanini:
We will implement the new additions at another time. See #836

PR #841, opened 2025-10-01
use agent name as branch name

treblereel:
I propose changing the branch naming when executing agentic parallel tasks.
Since the agent’s name will be used (or the method name if the agent’s name is not specified), the naming of execution results in the context will be strictly deterministic, which will significantly simplify further work with the results.

Drawbacks: if multiple instances of the same agent with identical names are used in a parallel task, the results of the later executions will overwrite the results of the earlier ones.

### DIFF
--- a/experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/AgentDslWorkflowTest.java
+++ b/experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/AgentDslWorkflowTest.java
@@ -101,7 +101,7 @@ class AgentDslWorkflowTest {
   @Test
   void dslParallelAgents() {
     var a1 = AgentsUtils.newMovieExpert();
-    var a2 = AgentsUtils.newMovieExpert();
+    var a2 = AgentsUtils.newMovieExpert2();
 
     Workflow wf = workflow("forkFlow").parallel("fanout", a1, a2).build();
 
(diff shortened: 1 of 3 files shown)


experimental/fluent/agentic/src/test/java/io/serverlessworkflow/fluent/agentic/AgentDslWorkflowTest.java
```diff
@@ -112,8 +112,8 @@ void dslParallelAgents() {
     // two branches created
     assertThat(fork.getFork().getBranches()).hasSize(2);
     // branch names follow "branch-{index}-{name}"
-    assertThat(fork.getFork().getBranches().get(0).getName()).isEqualTo("branch-0-fanout");
-    assertThat(fork.getFork().getBranches().get(1).getName()).isEqualTo("branch-1-fanout");
+    assertThat(fork.getFork().getBranches().get(0).getName()).isEqualTo("findMovie");
+    assertThat(fork.getFork().getBranches().get(1).getName()).isEqualTo("findMovie2");
```
ricardozanini, line 116: This is fine, but why not `findMovie-branch-0` and `findMovie-branch-1`?

treblereel: well, the findMovie comes MovieExpert.findMovie, because newMovieExpert() {} doesn't decalre name()
findMovie2 comes from newMovieExpert2() where newMovieExpert2() declares .name("findMovie2")

If you want I can fix it

ricardozanini: ok, I see. You are using the task's name as the branch name internally?

treblereel: yeap, thats it. And from my point of view, it's nice, coz enduser can set the name of the agent in the parallel loop and later do something with the output knowing which branch produces which result

ricardozanini (approved):

PR #842, opened 2025-10-01
Chained method calls should accumulate state rather than overwrite pr…

treblereel:
…evious values

### DIFF
--- a/experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentWorkflowBuilder.java
+++ b/experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentWorkflowBuilder.java
@@ -17,6 +17,7 @@ package io.serverlessworkflow.fluent.agentic;
 
 import dev.langchain4j.agentic.scope.AgenticScope;
 import io.serverlessworkflow.api.types.TaskItem;
+import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.fluent.func.spi.FuncTransformations;
 import io.serverlessworkflow.fluent.spec.BaseWorkflowBuilder;
 import java.util.ArrayList;
@@ -29,8 +30,11 @@ public class AgentWorkflowBuilder
     extends BaseWorkflowBuilder<AgentWorkflowBuilder, AgentDoTaskBuilder, AgentTaskItemListBuilder>
     implements FuncTransformations<AgentWorkflowBuilder> {
 
+  private final List<AgentDoTaskBuilder> agentDoTaskBuilders;
+
   AgentWorkflowBuilder(final String name, final String namespace, final String version) {
     super(name, namespace, version);
+    agentDoTaskBuilders = new ArrayList<>();
   }
 
   public static AgentWorkflowBuilder workflow() {
@@ -53,7 +57,7 @@ public class AgentWorkflowBuilder
   public AgentWorkflowBuilder agent(String name, Object agent) {
     final AgentDoTaskBuilder doTaskBuilder = this.newDo();
     doTaskBuilder.agent(name, agent);
-    setDo(doTaskBuilder);
+    agentDoTaskBuilders.add(doTaskBuilder);
     return this;
   }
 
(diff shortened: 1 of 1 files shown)


treblereel:
`Failed to execute goal org.codehaus.mojo:buildnumber-maven-plugin:3.2.1:create (get-scm-revision) on project serverlessworkflow-parent: Execution get-scm-revision of goal org.codehaus.mojo:buildnumber-maven-plugin:3.2.1:create failed: Plugin org.codehaus.mojo:buildnumber-maven-plugin:3.2.1 or one of its dependencies could not be resolved:
Error:  	Could not transfer artifact net.java.dev.jna:jna-platform:jar:5.6.0 from/to central (https://repo.maven.apache.org/maven2): Connection reset
Error:  -> [Help 1]
Error:  
Error:  To see the full stack trace of the errors, re-run Maven with the -e switch.
Error:  Re-run Maven using the -X switch to enable full debug logging.
Error:  
Error:  For more information about the errors and possible solutions, please read the following articles:
Error:  [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginResolutionException
Error: Process completed with exit code 1.`

hmm

experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentWorkflowBuilder.java
```diff
@@ -96,4 +100,14 @@ protected AgentDoTaskBuilder newDo() {
...
   }
+
+  private void setDo(AgentDoTaskBuilder doTaskBuilder) {
+    if (workflow.getDo() != null) {
+      List<TaskItem> existing = new ArrayList<>(workflow.getDo());
+      existing.addAll(doTaskBuilder.build().getDo());
+      this.workflow.setDo(Collections.unmodifiableList(existing));
+    } else {
+      this.workflow.setDo(doTaskBuilder.build().getDo());
+    }
```
fjtirado: Sure Im missing something but why not just do 

workflow.getDo().addAll(0, doTaskBuilder.build().getDo())?

As far as taskBuilder does not generate an inmutable list, you save a lot of objects.

treblereel: Because it's unmodifiableList

fjtirado: yes, Im asking to change the fact that this list is unmodifiable [here](https://github.com/serverlessworkflow/sdk-java/blob/main/fluent/spec/src/main/java/io/serverlessworkflow/fluent/spec/BaseTaskItemListBuilder.java#L70)

treblereel: yeap, that even better! I assumed that you would be against this, but since you yourself suggested it ...

fjtirado: To be honest, I do not understand why do is initialized several times, but the way it is done now, ir order to minimize changes, is better to not make it inmutable

fjtirado: Maybe you can try adding the task to the taskbuilder and calling do in the build of the end, I think the issue was calling setDo too early

treblereel: @fjtirado workflow.getDo().addAll(0, doTaskBuilder.build().getDo())?

by the way, index 0 means we insert before existing elms, I think we should add elms after

fjtirado: yes, I read your change badly and thought that you were adding new iteams at the beginning, but that should not be the case, so lets used the not indexed add

PR #864, opened 2025-10-08
[Fix #863] Adding WorkflowContext and TaskContext as additional params

### LINKED ISSUE #863: Allow passing WorkflowContext and TaskContext to input, output, export function for experimental labmda
Current these functions only accept the current model. In order to achieve their purpose they might need additional information (as it happen when using JQ). This information is contained in WorkflowContext and TaskContext, which might be optional extra parameters if needed

fjtirado:
User can now specify java filter functions that accepts WorkflowContext and TaskContext in their signature.

Fix https://github.com/serverlessworkflow/sdk-java/issues/863

### DIFF
--- a/experimental/lambda/src/test/java/io/serverless/workflow/impl/ModelTest.java
+++ b/experimental/lambda/src/test/java/io/serverless/workflow/impl/ModelTest.java
@@ -30,6 +30,7 @@ import io.serverlessworkflow.api.types.WaitTask;
 import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.api.types.func.OutputAsFunction;
 import io.serverlessworkflow.impl.WorkflowApplication;
+import io.serverlessworkflow.impl.WorkflowInstance;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.ExecutionException;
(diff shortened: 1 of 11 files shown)


experimental/lambda/src/test/java/io/serverless/workflow/impl/ModelTest.java
```diff
@@ -173,4 +174,62 @@ void testPOJOStringExpression() throws InterruptedException, ExecutionException
...
+                      new TaskItem(
+                          "doNothing",
+                          new Task()
+                              .withWaitTask(
+                                  new WaitTask()
+                                      .withWait(
+                                          new TimeoutAfter()
+                                              .withDurationInline(
+                                                  new DurationInline().withMilliseconds(10)))))))
+              .withOutput(
```
fjtirado, line 196: Output associated to a workflow, we specify the two arguments function.
Important thing to know is that TaskContext will be null if the three parameter function has been specified here (we should either document that or prevent user passing the three parameter function if the input/output/export is associated to the workflow)

experimental/lambda/src/test/java/io/serverless/workflow/impl/ModelTest.java
```diff
@@ -173,4 +174,62 @@ void testPOJOStringExpression() throws InterruptedException, ExecutionException
...
+                  List.of(
+                      new TaskItem(
+                          "doNothing",
+                          new Task()
+                              .withWaitTask(
+                                  new WaitTask()
+                                      .withOutput(
+                                          new Output()
+                                              .withAs(
+                                                  new OutputAsFunction()
```
fjtirado, line 223: Output associated to the task, we call three parameter function without issue
We can also call the two/one parameter one if we want.

experimental/types/src/main/java/io/serverlessworkflow/api/types/func/ExportAsFunction.java
```diff
@@ -31,4 +31,24 @@ public <T, V> ExportAs withFunction(Function<T, V> value, Class<T> argClass) {
     setObject(new TypedFunction<>(value, argClass));
     return this;
   }
+
```
fjtirado, line 34: Export also existed ;)

Issue #892, opened 2025-10-21
Add Shell process implementation


mcruzdev:
Implement [runShell](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#shell-process).

PR #896, opened 2025-10-22
add RunTask.workflow support

treblereel:
Fix https://github.com/serverlessworkflow/sdk-java/issues/897

This change introduces execution support for sub-workflows within the main workflow definition.
It allows a workflow to invoke another workflow declaratively via the RunTask construct

### DIFF
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/SubWorkflowTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/SubWorkflowTest.java
@@ -1,0 +1,146 @@
+/*
+ * Copyright 2020-Present The Serverless Workflow Specification Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.impl.test;
+
+import static io.serverlessworkflow.api.WorkflowReader.readWorkflowFromClasspath;
+import static org.junit.Assert.assertEquals;
+
+import io.serverlessworkflow.api.types.Workflow;
+import io.serverlessworkflow.impl.WorkflowApplication;
+import java.io.IOException;
+import java.util.Map;
+import org.junit.Test;
+
+public class SubWorkflowTest {
+
+  private static final String WORKFLOW_TEST_PATH = "workflows-samples/sub-workflow/";
+
+  @Test
+  public void setTest() throws IOException {
+    Workflow workflowParent =
+        readWorkflowFromClasspath(WORKFLOW_TEST_PATH + "sub-workflow-parent.yaml");
+    Workflow workflowChild =
+        readWorkflowFromClasspath(WORKFLOW_TEST_PATH + "sub-workflow-child.yaml");
+
+    try (WorkflowApplication app = WorkflowApplication.builder().build()) {
+      app.workflowDefinition(workflowChild);
+      Map<String, Object> result =
+          app.workflowDefinition(workflowParent)
+              .instance(Map.of())
+              .start()
+              .get()
+              .asMap()
+              .orElseThrow();
+      assertEquals("1", result.get("counter").toString());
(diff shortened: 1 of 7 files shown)


mcruzdev:
Great @treblereel I was working on `RunTask.shell` support  #892, I will wait this pull request to get your changes.

ricardozanini:
@treblereel can you please update the description and link the issue?

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunTaskExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+    protected RunWorkflowExecutor(RunTaskExecutorBuilder builder) {
+      super(builder);
+      this.application = builder.application;
+    }
+
+    @Override
+    protected CompletableFuture<WorkflowModel> internalExecute(
+        WorkflowContext workflowContext, TaskContext taskContext) {
+      if (taskContext.task() != null && taskContext.task() instanceof RunTask runTask) {
+        RunWorkflow runWorkflow = runTask.getRun().getRunWorkflow();
```
ricardozanini: This can be null since `getRun()` will have `scripts`, `shell`, `container`:

```java
    private RunTaskConfiguration value;
    private RunContainer runContainer;
    private RunScript runScript;
    private RunShell runShell;
    private RunWorkflow runWorkflow;
```

treblereel: @ricardozanini I agree this could be revisited later, but for now, the scope of this executor is limited to run.workflow Once additional Run* types are introduced, we can generalize the logic and handle nulls consistently across all cases

fjtirado (changes requested):
I think RunTaskExecutor should mimic CallTaskExecutor. 

so rather than directly instantiating RunWorkflowExceutor, load the four different executor using serviceloader. 

The reason is that the other run task will require additional dependencies and should be located in different modules (like HTTP or OpenAPI) 

So, even if running workflow does not need additional dependencies, it should be loaded through serivce loading mechanism

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunTaskExecutor.java
fjtirado: I think this should be done first, not later.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunTaskExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+              && workflowDefinitionId.namespace().equals(namespace)
+              && workflowDefinitionId.version().equals(version)) {
+            WorkflowDefinition workflowDefinition = kv.getValue();
+            WorkflowApplication workflowApplication = workflowDefinition.application();
+            return workflowApplication
+                .workflowDefinition(workflowDefinition.workflow())
+                .instance(taskContext.input().asJavaObject())
+                .start();
+          }
+        }
```
fjtirado: WorkflowDefinition identification does not need to be done every time the workflow is executed, most of this code should be executed just once.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunTaskExecutor.java
```diff
@@ -0,0 +1,95 @@
...
+            application.workflowDefinitions().entrySet()) {
+          WorkflowDefinitionId workflowDefinitionId = kv.getKey();
+          if (workflowDefinitionId.name().equals(name)
+              && workflowDefinitionId.namespace().equals(namespace)
+              && workflowDefinitionId.version().equals(version)) {
+            WorkflowDefinition workflowDefinition = kv.getValue();
+            WorkflowApplication workflowApplication = workflowDefinition.application();
+            return workflowApplication
+                .workflowDefinition(workflowDefinition.workflow())
+                .instance(taskContext.input().asJavaObject())
```
fjtirado: you already have the workflowDefinition, you do not need to retreive it again from the appliaction. 
Also, you can create a WorkflowDefinitionId is a record, the proper way to do this should be

```
WorkflowDefinition workflowDefinition = application.workflowDefinitions().get(new WorkflowDefinitionId(name, namespace.version));
if (workflowDefinition != null) {

                 return workflowDefinition.instance(taskContext.input()).start();
} else {
   // throw exception indicating the workflow is not defined
}

```

Also, you do not need to call asJavaObject(), instance should accept a WorkflowModel as per https://github.com/serverlessworkflow/sdk-java/blob/main/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowModelFactory.java#L72-L74

fjtirado (changes requested):
I think we need to refactor the PR once I have fixed https://github.com/serverlessworkflow/sdk-java/issues/899

fjtirado (approved):

PR #898, opened 2025-10-23
Add RunTask.shell task

### LINKED ISSUE #892: Add Shell process implementation

mcruzdev:
## Context

Implement [RunTask.shell](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#shell-process) task:

```yaml
document:
  dsl: '1.0.1'
  namespace: test
  name: run-shell-example
  version: '0.1.0'
do:
  - runShell:
      run:
        shell:
          command: 'echo "Hello, ${ .user.name }"'
```

## Tracking

- [x] Add arguments
- [x] Implement `await`
- [x] Implement `return`
- [ ] [Support multiple OS 
](https://github.com/serverlessworkflow/sdk-java/pull/898/files#diff-a85b99dc6ddc3f55cff23fedaf608d3a55833cfa86700b68d86c81b726a22729R89)

Closes #892

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
@@ -28,16 +28,10 @@ import io.serverlessworkflow.impl.WorkflowModel;
 import io.serverlessworkflow.impl.WorkflowModelFactory;
 import io.serverlessworkflow.impl.WorkflowUtils;
 import io.serverlessworkflow.impl.expressions.ExpressionUtils;
-import java.io.BufferedReader;
 import java.io.IOException;
-import java.io.InputStream;
-import java.io.InputStreamReader;
-import java.io.StringWriter;
-import java.io.UncheckedIOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.Supplier;
 
 public class RunShellExecutor implements RunnableTask<RunShell> {
 
@@ -59,9 +53,8 @@ public class RunShellExecutor implements RunnableTask<RunShell> {
   public CompletableFuture<WorkflowModel> apply(
       WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input) {
     ProcessBuilder processBuilder = this.processBuilderSupplier.apply(workflowContext, taskContext);
-    WorkflowModel workflowModel =
-        this.shellResultSupplier.apply(taskContext, input, processBuilder);
-    return CompletableFuture.completedFuture(workflowModel);
+    return CompletableFuture.completedFuture(
+        this.shellResultSupplier.apply(taskContext, input, processBuilder));
   }
 
   @Override
@@ -145,34 +138,29 @@ public class RunShellExecutor implements RunnableTask<RunShell> {
             Process process = processBuilder.start();
 
             if (taskConfiguration.isAwait()) {
-              return waitForResult(taskConfiguration, definition, process);
+              return buildResultFromProcess(taskConfiguration, definition, process);
             } else {
               return input;
             }
 
-          } catch (IOException | InterruptedException | UncheckedIOException e) {
+          } catch (IOException | InterruptedException e) {
             throw new WorkflowException(WorkflowError.runtime(taskContext, e).build(), e);
           }
         };
   }
 
-  private WorkflowModel waitForResult(
+  /**
+   * Builds the WorkflowModel result from the executed process. It waits for the process to finish
+   * and captures the exit code, stdout, and stderr based on the task configuration.
+   */
+  private WorkflowModel buildResultFromProcess(
       RunShell taskConfiguration, WorkflowDefinition definition, Process process)
       throws IOException, InterruptedException {
 
-    CompletableFuture<String> futureStdout =
-        CompletableFuture.supplyAsync(inputStreamStringSupplier(process.getInputStream()));
-    CompletableFuture<String> futureStderr =
-        CompletableFuture.supplyAsync(inputStreamStringSupplier(process.getErrorStream()));
-
     int exitCode = process.waitFor();
 
-    CompletableFuture<Void> allStd = CompletableFuture.allOf(futureStdout, futureStderr);
-
-    allStd.join();
-
-    String stdout = futureStdout.join();
-    String stderr = futureStderr.join();
+    String stdout = new String(process.getInputStream().readAllBytes(), StandardCharsets.UTF_8);
+    String stderr = new String(process.getErrorStream().readAllBytes(), StandardCharsets.UTF_8);
 
     RunTaskConfiguration.ProcessReturnType returnType = taskConfiguration.getReturn();
 
@@ -187,38 +175,8 @@ public class RunShellExecutor implements RunnableTask<RunShell> {
     };
   }
 
-  private static Supplier<String> inputStreamStringSupplier(InputStream process) {
-    return () -> {
-      try {
-        return readInputStream(process);
-      } catch (IOException e) {
-        throw new UncheckedIOException(e);
-      }
-    };
-  }
-
   @Override
   public boolean accept(Class<? extends RunTaskConfiguration> clazz) {
     return RunShell.class.equals(clazz);
-  }
-
-  /**
-   * Reads an InputStream and returns its content as a String. It keeps the original content using
-   * UTF-8 encoding.
-   *
-   * @param inputStream {@link InputStream} to be read
-   * @return {@link String} with the content of the InputStream
-   */
-  private static String readInputStream(InputStream inputStream) throws IOException {
-    StringWriter writer = new StringWriter();
-    try (BufferedReader reader =
-        new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
-      char[] buffer = new char[1024];
-      int charsRead;
-      while ((charsRead = reader.read(buffer)) != -1) {
-        writer.write(buffer, 0, charsRead);
-      }
-    }
-    return writer.toString();
   }
 }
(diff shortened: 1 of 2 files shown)


mcruzdev:
@ricardozanini could you share with me a example of arguments?  Should be something similar to: 

```yaml
document:
  dsl: '1.0.1'
  namespace: test
  name: run-shell-example
  version: '0.1.0'
do:
  - runShell:
      run:
        shell:
          command: 'kubectl'
          arguments:
            - 'apply'
            - '-f'
            - '/k8s'
```

And the output should be:

```shell
kubectl apply -f /k8s
```

I am not sure about it, do you have a example?

fjtirado:
@mcruzdev I have merged https://github.com/serverlessworkflow/sdk-java/pull/900, please rebase and continue from there, thanks

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,162 @@
...
+
+public class RunShellExecutor implements RunnableTask<RunShell> {
+
+  private ShellResultSupplier shellResultSupplier;
+  private ProcessBuilderSupplier processBuilderSupplier;
+
+  @FunctionalInterface
+  private interface ShellResultSupplier {
+    WorkflowModel apply(
+        TaskContext taskContext, ProcessBuilder processBuilder, WorkflowModel input);
```
fjtirado: Really picky comment, but I was thinking that since oter interface has WorkflowContext, TaskContext, and WorkflowModel in that order, I thing the order here, should probably be TaskContext taskContext, WorkflowModel input and ProcessBuilder processBuilder.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,162 @@
...
+  }
+
+  private interface ProcessBuilderSupplier {
+    ProcessBuilder apply(TaskContext taskContext, WorkflowContext workflowContext);
```
fjtirado: Same here, probably we should switch the parameter order to be WorkflowContext and TaskContext.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,162 @@
...
+
+          return builder;
+        };
+
+    this.shellResultSupplier =
+        (taskContext, processBuilder, input) -> {
+          if (shellCommand == null || shellCommand.isBlank()) {
+            throw new IllegalStateException(
+                "Missing shell command in RunShell task: " + taskContext.taskName());
+          }
```
fjtirado: Maybe this check can be done outside the supplier, I mean if the command is null or empty, the definition is wrong and should be corrected

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,162 @@
...
+            Process process = processBuilder.start();
+
+            if (taskConfiguration.isAwait()) {
+              return executeAwaiting(taskConfiguration, definition, process);
+            } else {
+              return input;
+            }
+
+          } catch (IOException | InterruptedException e) {
+            throw new RuntimeException(e);
```
fjtirado: Hmm, here we should throw a WorkflowException with a runtime error https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#standard-error-types
This way the error can be handled by a TryTask

mcruzdev:
> > > @mcruzdev yes, the arguments are a list: https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#shell-process
> > > Which kind of example do you need?
> > 
> > 
> > Ok, it is all, I think I do not need a example 😄
> > I had a confusion due to `map` as type of `arguments` in the [documentation](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#properties-17).
> 
> You can do:
> 
> ```yaml
> arguments:
>   - key: value
> ```
> 
> Maybe in some use cases that would make sense.

It is not clear to me again @ricardozanini, what is the real command statement in that case? The arguments will be used like an input ([like `script` does](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#examples-15))?  Or we will have something like `customcommand key=value` or `customcommand key=value` in the statement.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,162 @@
...
+          }
+        };
+  }
+
+  private WorkflowModel executeAwaiting(
+      RunShell taskConfiguration, WorkflowDefinition definition, Process process)
+      throws IOException, InterruptedException {
+    StringBuilder stdout = new StringBuilder();
+    StringBuilder stderr = new StringBuilder();
```
fjtirado: Althour the override of outputstream is quite imaginative and I like it, I have to mention that is better to use standard library, using
```

StringWriter outWriter = new StringWriter();
process.getInputStream().transferTo(new OutputStreamWriter(outWriter));

.....

outWriter.toString();
```

mathecruz: I think that the `OutputStreamWriter` is not possible, but yes, we need a better solution that is not a byte to byte.

EDIT: I mean in the `transferTo` method.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,214 @@
...
+  }
+
+  /**
+   * Reads an InputStream and returns its content as a String. It keeps the original content using
+   * UTF-8 encoding.
+   *
+   * @param inputStream {@link InputStream} to be read
+   * @return {@link String} with the content of the InputStream
+   */
+  public static String readInputStream(InputStream inputStream) {
```
fjtirado: I think this method should be private, it is not really used outside this class

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,214 @@
...
+    StringWriter writer = new StringWriter();
+    try (BufferedReader reader =
+        new BufferedReader(new InputStreamReader(inputStream, StandardCharsets.UTF_8))) {
+      char[] buffer = new char[1024];
+      int charsRead;
+      while ((charsRead = reader.read(buffer)) != -1) {
+        writer.write(buffer, 0, charsRead);
+      }
+    } catch (IOException e) {
```
fjtirado: Let the IOException being thrown so it is catched by internalExecute and convert to a proper workflow error

fjtirado (commented):
@mcruzdev Another coupple of comments related with IOException handling

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,214 @@
...
+    ProcessBuilder apply(WorkflowContext workflowContext, TaskContext taskContext);
+  }
+
+  @Override
+  public CompletableFuture<WorkflowModel> apply(
+      WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input) {
+    ProcessBuilder processBuilder = this.processBuilderSupplier.apply(workflowContext, taskContext);
+    WorkflowModel workflowModel =
+        this.shellResultSupplier.apply(taskContext, input, processBuilder);
+    return CompletableFuture.completedFuture(workflowModel);
```
fjtirado: Probably this should be called asynchronusly  ( I added more comments to other places affected by this changed)
```suggestion
  ProcessBuilder processBuilder = this.processBuilderSupplier.apply(workflowContext, taskContext);
   return CompletableFuture.supplyAsync(() -> shellResultSupplier.apply(taskContext, input,processBuilder));
```

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,224 @@
...
+    };
+  }
+
+  private static Supplier<String> inputStreamStringSupplier(InputStream process) {
+    return () -> {
+      try {
+        return readInputStream(process);
+      } catch (IOException e) {
+        throw new UncheckedIOException(e);
+      }
```
fjtirado: When this is change back to return string, lets just throw IOException, you were already catching it on the main block

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,224 @@
...
+            throw new WorkflowException(WorkflowError.runtime(taskContext, e).build(), e);
+          }
+        };
+  }
+
+  private WorkflowModel waitForResult(
+      RunShell taskConfiguration, WorkflowDefinition definition, Process process)
+      throws IOException, InterruptedException {
+
+    CompletableFuture<String> futureStdout =
```
fjtirado: If you make async the whole procedure, you do not need this to be async

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -0,0 +1,224 @@
...
+          try {
+            Process process = processBuilder.start();
+
+            if (taskConfiguration.isAwait()) {
+              return waitForResult(taskConfiguration, definition, process);
+            } else {
+              return input;
+            }
+
+          } catch (IOException | InterruptedException | UncheckedIOException e) {
```
fjtirado: if inputStreamSupplier is not a supplier (its not longer async because the whole execution is performed asynchronusly), then you do not need to catch UncheckedIOException here)

PR #910, opened 2025-10-24
[Fix #898] RunShell minor improvements

fjtirado:
Enable test only in linux (till we implement for other Operating systems)
Run asynchronously

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
@@ -53,8 +53,8 @@ public class RunShellExecutor implements RunnableTask<RunShell> {
   public CompletableFuture<WorkflowModel> apply(
       WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input) {
     ProcessBuilder processBuilder = this.processBuilderSupplier.apply(workflowContext, taskContext);
-    return CompletableFuture.completedFuture(
-        this.shellResultSupplier.apply(taskContext, input, processBuilder));
+    return CompletableFuture.supplyAsync(
+        () -> this.shellResultSupplier.apply(taskContext, input, processBuilder));
   }
 
   @Override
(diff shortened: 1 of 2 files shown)


impl/test/src/test/java/io/serverlessworkflow/impl/test/RunShellExecutorTest.java
```diff
@@ -152,8 +155,7 @@ void testStderr() throws IOException {
           softly -> {
             softly.assertThat(outputModel.asText()).isPresent();
             softly.assertThat(outputModel.asText().get()).isNotEmpty();
-            softly.assertThat(outputModel.asText().get()).contains("ls: cannot access");
-            softly.assertThat(outputModel.asText().get()).contains("No such file or directory");
+            softly.assertThat(outputModel.asText().get()).contains("ls");
```
fjtirado: This test only worked if the locale of the OS was english, so I do the trick of just checking that the stderr contains "ls"

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -128,21 +124,16 @@ public void init(RunShell taskConfiguration, WorkflowDefinition definition) {
...
           try {
             Process process = processBuilder.start();
-
-            if (taskConfiguration.isAwait()) {
-              return buildResultFromProcess(taskConfiguration, definition, process);
-            } else {
-              return input;
-            }
-
+            return taskConfiguration.isAwait()
```
fjtirado, line 134: yes, this one was not needed, but I love ternary operator and could not refrain, sorry ;)

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -69,48 +69,44 @@ public void init(RunShell taskConfiguration, WorkflowDefinition definition) {
...
-                  ExpressionUtils.isExpr(entry.getKey())
-                      ? WorkflowUtils.buildStringResolver(
-                              application, entry.getKey(), taskContext.input().asJavaObject())
-                          .apply(workflowContext, taskContext, taskContext.input())
-                      : entry.getKey();
-
-              if (entry.getValue() == null) {
-                commandBuilder.append(" ").append(argKey);
-                continue;
+              commandBuilder
```
fjtirado, line 84: I wanted to avoid the use of continue, which in this case was possible

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -69,48 +69,44 @@ public void init(RunShell taskConfiguration, WorkflowDefinition definition) {
...
 
-          String command =
-              ExpressionUtils.isExpr(shellCommand)
-                  ? WorkflowUtils.buildStringResolver(
-                          application, shellCommand, taskContext.input().asJavaObject())
-                      .apply(workflowContext, taskContext, taskContext.input())
-                  : shellCommand;
-
-          StringBuilder commandBuilder = new StringBuilder(command);
+          StringBuilder commandBuilder =
```
fjtirado, line 72: Just saving a variable declaration

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
mathecruz: 😆

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -53,8 +53,8 @@ private interface ProcessBuilderSupplier {
   public CompletableFuture<WorkflowModel> apply(
       WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input) {
     ProcessBuilder processBuilder = this.processBuilderSupplier.apply(workflowContext, taskContext);
-    return CompletableFuture.completedFuture(
-        this.shellResultSupplier.apply(taskContext, input, processBuilder));
+    return CompletableFuture.supplyAsync(
```
mathecruz, line 56: I forgot about that, nice

PR #927, opened 2025-10-30
[Fix #925] Allow association of additional objects to application

### LINKED ISSUE #925: Allow passing extension objects to WorkflowApplication
In order to further extent implementation customization capabilities, it will be good to be able to associate arbitrary objects to the application, so it can be later accessed. 

For example, default http imple use JaxRs client (itself a factory), but it might be handy for some implementors to replace it. Since we do not want to create a hard dependency between the core (where WorkflowApplication is) and jaxrs or any other rest client, the client object might be added by the implementor when building the application and later retrieved in the task implementation class

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/925

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
@@ -19,6 +19,7 @@ import static io.serverlessworkflow.impl.WorkflowUtils.safeClose;
 
 import io.serverlessworkflow.api.types.SchemaInline;
 import io.serverlessworkflow.api.types.Workflow;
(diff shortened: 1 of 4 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -221,6 +225,14 @@ public Builder withEventPublisher(EventPublisher eventPublisher) {
       return this;
     }
 
+    public <T> Builder withExtensionObject(String name, WorkflowAdditionalObject additionalObject) {
```
ricardozanini: I'd rename this to `withAdditionalObject` for ubiquity if the other methods.

ricardozanini (approved):

PR #940, opened 2025-10-31
[Fix #930] Adding authorization runtime expression

### LINKED ISSUE #930: Implement authorization runtime expression
Implement `$authorization` in runtime expressions

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/930

This also fixed following things:
- Secret should be a map, not a string
- Expression exception should be a workflow exception of the proper type as per spec, not an IllegalArgument one
- The AdditionaObject factory should accept WorkflowContextData and TaskContextData, so it has not access to internals exposed by their parent interfaces.

### DIFF
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OAuth2AuthProvider.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OAuth2AuthProvider.java
@@ -22,17 +22,12 @@ import io.serverlessworkflow.impl.TaskContext;
 import io.serverlessworkflow.impl.WorkflowApplication;
 import io.serverlessworkflow.impl.WorkflowContext;
 import io.serverlessworkflow.impl.WorkflowModel;
-import io.serverlessworkflow.impl.executors.http.auth.jwt.JWT;
 import io.serverlessworkflow.impl.executors.http.auth.requestbuilder.AuthRequestBuilder;
 import io.serverlessworkflow.impl.executors.http.auth.requestbuilder.OAuthRequestBuilder;
-import jakarta.ws.rs.client.Invocation;
-import jakarta.ws.rs.client.Invocation.Builder;
 
-public class OAuth2AuthProvider implements AuthProvider {
+public class OAuth2AuthProvider extends AbstractAuthProvider {
 
   private AuthRequestBuilder requestBuilder;
-
-  private static final String BEARER_TOKEN = "Bearer %s";
 
   public OAuth2AuthProvider(
       WorkflowApplication application, Workflow workflow, OAuth2AuthenticationPolicy authPolicy) {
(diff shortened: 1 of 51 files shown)


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OAuth2AuthProvider.java
```diff
@@ -46,15 +41,12 @@ public OAuth2AuthProvider(
...
+    return requestBuilder.build(workflow, task, model).validateAndGet().token();
   }
 
   @Override
-  public void preRequest(
-      Invocation.Builder builder, WorkflowContext workflow, TaskContext task, WorkflowModel model) {
-    JWT jwt = requestBuilder.build(workflow, task, model).validateAndGet();
-    builder.header(AuthProviderFactory.AUTH_HEADER_NAME, String.format(BEARER_TOKEN, jwt.token()));
+  protected String authScheme() {
+    return "Bearer";
```
ricardozanini, line 50: Is this correct? This is the `OAuth2Provider`

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OpenIdAuthProvider.java
```diff
@@ -50,15 +45,12 @@ public OpenIdAuthProvider(
...
+    return requestBuilder.build(workflow, task, model).validateAndGet().token();
   }
 
   @Override
-  public void preRequest(
-      Invocation.Builder builder, WorkflowContext workflow, TaskContext task, WorkflowModel model) {
-    JWT jwt = requestBuilder.build(workflow, task, model).validateAndGet();
-    builder.header(AuthProviderFactory.AUTH_HEADER_NAME, String.format(BEARER_TOKEN, jwt.token()));
+  protected String authScheme() {
+    return "Bearer";
```
ricardozanini, line 54: Same thing as the OAuth2 provider. Probably you can do an OpenID -> OAuth2 inheritance since conceptually Oauth2 derives from openID.

ricardozanini (commented):
Also, do we already have tests for OAuth2? I see you added for Basic only.

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OAuth2AuthProvider.java
fjtirado: It also surprised me, but the fact is that in the the authorization header that is added after all communication with the oauth server, what is in there is named bearer (which makes sense  since at the end the result of all the calculation is the bearer token). 
Note that having $authorization  allows saving it to the context for internal reuse without really exposing it ($authorization.parameter) and $authorization.scheme allows you to check if the value is a bearer token.

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/OpenIdAuthProvider.java
fjtirado: yes, the inheritance from the same abstract class is already there, I opened https://github.com/serverlessworkflow/sdk-java/issues/952 for deeper refactor

PR #941, opened 2025-11-01
chore: fix CI by removing examples module; remove *.db files from tests module

ricardozanini:
**In this PR:**
- ~Correcting the sample workflow `set-listen-to-any.yaml` so a listen actually waits (not completes immediately due to `listen.any: []`).~
- ~Generating DB samples at test time (under `@TempDir`) instead of committing binary DB files.~
- ~Making the restore tests parameterized.~
- ~Adding deterministic awaits in the DB generator to eliminate race conditions.~
- Remove `examples` from release mvn command.

**Follow up:**
- ~To check if the desired behavior is to finish the workflow immediately on `listen.to.any: []`. I believe that's the desired behavior, but it's worth checking out.~

~Not sure how the `DbGenerator` was working before, since the sample workflow was completed immediately/having a racing condition before suspending/finishing the workflow by listening to anything empty.~

~I created a small/raw/workaround to run the generator on a separate JVM when running the tests to avoid a collision with the file read. It's working pretty good, if you have a better idea (or any changes in the store implementation to avoid this), let me know.~

### DIFF
--- a/impl/test/pom.xml
+++ b/impl/test/pom.xml
@@ -117,7 +117,7 @@
                                     <goal>java</goal>
                                 </goals>
                                 <configuration>
-                                    <mainClass>io.serverlessworkflow.impl.test.DBGeneratorCli</mainClass>
+                                    <mainClass>io.serverlessworkflow.impl.test.DBGenerator</mainClass>
                                 </configuration>
                             </execution>
                         </executions>


fjtirado (commented):
I do not get why storing binary file is worse than spawning a new JVM.
Also the DB files were stored on purpose to check backward compatibility with previous DB versions. They are intentionally NOT generated with every run

impl/test/db-samples/running.db
fjtirado, line 1: This file was generated using a previous version of the writer. It should NOT be removed

impl/test/src/test/java/io/serverlessworkflow/impl/test/MvStorePersistenceTest.java
```diff
@@ -27,31 +27,59 @@
...
+    ForkedDbGen.run(runningV1, false);
+    ForkedDbGen.run(suspendedV1, true);
+    ForkedDbGen.run(runningV0, false);
+    ForkedDbGen.run(suspendedV0, true);
+  }
 
   @Test
   void testSimpleRun() throws IOException {
-    final String dbName = "db-samples/simple.db";
+    final Path dbPath = tmp.resolve("simple.db");
```
fjtirado: Current version of the db writer is tested here. 
The other test were designed to test well known version of the DBs, thats why they should not be generated in every test

impl/test/src/test/resources/workflows-samples/set-listen-to-any.yaml
```diff
@@ -10,4 +10,6 @@ do:
   - callDoctor:
       listen:
         to:
-          any: []
```
fjtirado, line 13: This test was intenationally specyfing[] to test `EventConsumer.registerAll` method, it should not be associated with an specific type

impl/test/src/test/java/io/serverlessworkflow/impl/test/DBGenerator.java
```diff
@@ -16,27 +16,25 @@
...
 import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
+import java.time.Duration;
 import java.util.Map;
 
-public class DBGenerator {
+public final class DBGenerator {
 
-  public static void main(String[] args) throws IOException {
```
fjtirado, line 34: I think this PR should just commented the main

impl/test/db-samples/suspended.db
fjtirado, line 1: same here, this file was generated with a previous version of the code and should not be removed

impl/test/src/test/java/io/serverlessworkflow/impl/test/ForkedDbGen.java
```diff
@@ -0,0 +1,54 @@
...
+import java.io.*;
+import java.nio.file.*;
+import java.util.Optional;
+
+final class ForkedDbGen {
+  private static String detectJavaCmd() {
+    Optional<String> cmd = ProcessHandle.current().info().command();
+    if (cmd.isPresent() && Files.isRegularFile(Path.of(cmd.get()))) return cmd.get();
+    String home = System.getProperty("java.home");
+    String exe = System.getProperty("os.name").toLowerCase().contains("win") ? "java.exe" : "java";
```
fjtirado: I do not think this file is needed for this PR because I do not think we should run an extra JVM to generate the db binaries for every test run, but you can run executable in windows without appending the .exe, sou you do not need to check the os version to add the .exe in windows

impl/test/src/test/java/io/serverlessworkflow/impl/test/ForkedDbGen.java
```diff
@@ -0,0 +1,54 @@
...
+            javaCmd,
+            "-cp",
+            cp,
+            DBGeneratorCli.class.getCanonicalName(),
+            db.toString(),
+            String.valueOf(suspend));
+    pb.redirectErrorStream(true);
+
+    Process p = pb.start();
+    try (BufferedReader r = new BufferedReader(new InputStreamReader(p.getInputStream()))) {
```
fjtirado: The same can be achieved by 
`System.out.printl(new String (p.getInputStream().readAllBytes()))`
but since db generators does not really print anything int he console, I do not see the point of doing allo of this

impl/test/src/test/java/io/serverlessworkflow/impl/test/MvStorePersistenceTest.java
```diff
@@ -27,31 +27,59 @@
...
+  @TempDir static Path tmp;
+
+  static Path runningV1, suspendedV1, runningV0, suspendedV0;
+
+  @BeforeAll
+  static void prepareDbSamples() throws IOException, InterruptedException {
+    runningV1 = tmp.resolve("running_v1.db");
+    suspendedV1 = tmp.resolve("suspended_v1.db");
+    runningV0 = tmp.resolve("running.db");
+    suspendedV0 = tmp.resolve("suspended.db");
```
fjtirado: DB files were stored on the repo rather than generated on every run on purpose. Running and suspended are older versions that we want to test are still read by the code

impl/test/src/test/java/io/serverlessworkflow/impl/test/DBGenerator.java
```diff
@@ -51,8 +49,16 @@ private static void runInstance(String dbName, boolean suspend) throws IOExcepti
               readWorkflowFromClasspath("workflows-samples/set-listen-to-any.yaml"));
       WorkflowInstance instance = definition.instance(Map.of());
       instance.start();
+
+      await()
```
fjtirado: The start mehtod returns after the listener has been registered because event registration is a sync completable 
The suspend method return when the workflow has been effectiviely suspended. 
Thats why I did not add anything else here. 
Alss, this routine, is not a test, it does not make sense to use a testlibrary like awaibaility here.

PR #942, opened 2025-11-03
Add initial RunContainer Task support

treblereel:

### DIFF
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/ContainerTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/ContainerTest.java
@@ -45,7 +45,7 @@ import org.junit.jupiter.api.condition.DisabledIf;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
-@DisabledIf("checkDocker")
+@DisabledIf("isDockerNotAvailable")
 public class ContainerTest {
 
   private static DockerClient dockerClient;
@@ -61,7 +61,7 @@ public class ContainerTest {
   }
 
   @SuppressWarnings("unused")
-  private static boolean checkDocker() {
+  private static boolean isDockerNotAvailable() {
     try {
       dockerClient.pingCmd().exec();
       return false;


impl/container/src/main/java/io/serverlessworkflow/impl/container/executors/StringExpressionResolver.java
fjtirado, line 1: There is already an existing class that fulfills this purpose, 
`WorklowValueResolver<String>`

fjtirado (approved):
Great work!
Before merging I think @ricardozanini should review this container handling block https://github.com/serverlessworkflow/sdk-java/pull/942/files#diff-124b1b11448a7a123f2b5b8b30d8a10568d3cdbb69aed81169350425ee4ea070R139-R199

impl/container/pom.xml
```diff
@@ -0,0 +1,33 @@
...
+         xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+         xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+    <modelVersion>4.0.0</modelVersion>
+    <parent>
+        <groupId>io.serverlessworkflow</groupId>
+        <artifactId>serverlessworkflow-impl</artifactId>
+        <version>8.0.0-SNAPSHOT</version>
+    </parent>
+    <artifactId>serverlessworkflow-impl-container</artifactId>
+    <name>Serverless Workflow :: Impl :: OpenAPI</name>
```
ricardozanini:
```suggestion
    <name>Serverless Workflow :: Impl :: Container</name>
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/ContainerTest.java
```diff
@@ -0,0 +1,245 @@
...
+
+    String containerId = findContainerIdByName(containerName);
+    InspectContainerResponse inspect = dockerClient.inspectContainerCmd(containerId).exec();
+    Map<ExposedPort, Ports.Binding[]> ports = inspect.getNetworkSettings().getPorts().getBindings();
+
+    assertTrue(ports.containsKey(ExposedPort.tcp(8880)));
+    assertTrue(ports.containsKey(ExposedPort.tcp(8881)));
+    assertTrue(ports.containsKey(ExposedPort.tcp(8882)));
+
+    dockerClient.removeContainerCmd(containerId).withForce(true).exec();
```
ricardozanini: Can you add this to a finally block?

ricardozanini: The same applies to all tests; the procedure remains the same.

impl/test/src/test/resources/workflows-samples/container/container-cleanup-default.yaml
```diff
@@ -0,0 +1,12 @@
+document:
+  dsl: '1.0.2'
+  namespace: test
+  name: container-cleanup-default
+  version: '0.1.0'
+do:
+  - runContainer:
+      run:
+        container:
+          image: alpine:latest
```
ricardozanini: We can use `busybox` instead: https://hub.docker.com/_/busybox

It's less than 1MB.

fjtirado:
_Image allow-list / registry auth
Consider an allow-list of images/prefixes and registry credentials. Otherwise, anyone can run :latest from anywhere.
Recommendation: resolve images through a policy component; support private registry auth via AuthConfig.
We might have to add a config to the engine core such as sw.impl.containers.policy - @fjtirado_

I do not exactly see what the issue is, but in any case we already have a configuration [object](https://github.com/serverlessworkflow/sdk-java/blob/main/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java#L389) . 
If we want (because the IA mention it ;)) to define which images can be run on advance, we can add a property with the list of images allowed, but I do not see a real reason to do that and I do not see this affecting the core engine ;)

fjtirado:
_Singleton Docker client lifecycle
DockerClientHolder is static and never closed._

This is an intentional design choice. This is lazily loaded by the JVM the first time is used and keep alive till the process is done (since the WorkflowApplication should be tied with the proces lifecycle, it should be fine)
But.... we probably should allow users to be able to change the default client, like we are doing now with the Http client, take a look to https://github.com/serverlessworkflow/sdk-java/blob/main/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpClientResolver.java#L32-L37 and `WorkflowApplication.withAdditionalObject`

fjtirado:
_Comment says “must be removed because of withAutoRemove(true)” but I don’t see you setting withAutoRemove(true) here (maybe a property setter does it). If it’s not guaranteed, you might orphan containers._

Yes, a property setter is setting it when policy is ALWAYS or NEVER

impl/test/src/test/java/io/serverlessworkflow/impl/test/ContainerTest.java
```diff
@@ -0,0 +1,263 @@
...
+    DefaultDockerClientConfig defaultConfig =
+        DefaultDockerClientConfig.createDefaultConfigBuilder().build();
+    dockerClient =
+        DockerClientImpl.getInstance(
+            defaultConfig,
+            new ApacheDockerHttpClient.Builder().dockerHost(defaultConfig.getDockerHost()).build());
+  }
+
+  @SuppressWarnings("unused")
+  private static boolean checkDocker() {
```
ricardozanini:
```suggestion
  private static boolean isDockerNotAvailable() {
```

Then we would have:

```java
@DisabledIf("isDockerNotAvailable")
```

PR #962, opened 2025-11-04
Add support for Run.script (Javascript)

mcruzdev:

### DIFF
--- a/impl/pom.xml
+++ b/impl/pom.xml
@@ -16,6 +16,7 @@
         <version.org.glassfish.jersey>3.1.11</version.org.glassfish.jersey>
         <version.com.cronutils>9.2.1</version.com.cronutils>
         <version.docker.java>3.6.0</version.docker.java>
+        <version.org.graalvm.polyglot>23.1.1</version.org.graalvm.polyglot>
     </properties>
     <dependencyManagement>
         <dependencies>
@@ -132,11 +133,6 @@
                 <version>${version.com.h2database}</version>
             </dependency>
             <dependency>
-                <groupId>com.cronutils</groupId>
-                <artifactId>cron-utils</artifactId>
-                <version>${version.com.cronutils}</version>
-            </dependency>
-            <dependency>
                 <groupId>io.serverlessworkflow</groupId>
                 <artifactId>serverlessworkflow-impl-script-js</artifactId>
                 <version>${project.version}</version>
@@ -155,6 +151,17 @@
                 <groupId>com.github.docker-java</groupId>
                 <artifactId>docker-java-transport-httpclient5</artifactId>
                 <version>${version.docker.java}</version>
+            </dependency>
+            <dependency>
+                <groupId>org.graalvm.polyglot</groupId>
+                <artifactId>js</artifactId>
+                <version>${version.org.graalvm.polyglot}</version>
+                <type>pom</type>
+            </dependency>
+            <dependency>
+                <groupId>org.graalvm.polyglot</groupId>
+                <artifactId>polyglot</artifactId>
+                <version>${version.org.graalvm.polyglot}</version>
             </dependency>
         </dependencies>
     </dependencyManagement>
--- a/impl/script-js/pom.xml
+++ b/impl/script-js/pom.xml
@@ -5,9 +5,6 @@
         <artifactId>serverlessworkflow-impl</artifactId>
         <version>8.0.0-SNAPSHOT</version>
     </parent>
-    <properties>
-      <version.org.graalvm.polyglot>23.1.1</version.org.graalvm.polyglot>
-    </properties>
     <artifactId>serverlessworkflow-impl-script-js</artifactId>
     <name>Serverless Workflow :: Impl :: Script JavaScript</name>
     <dependencies>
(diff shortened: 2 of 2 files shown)


fjtirado (commented):
I will write a long general comment, since this requires some refactor. 
The goal should be to not add Javascript or Python dependencies to the core. 
With that spirit, you need a RunScriptTask in the core modules that selects either the JavaScript or the Python implementation (use the language id to select the right implementation, all implementation being loaded through service loading to not create dependencies)
Then you have another module with the java script specific code plus the javascript library (the java script implementation class being loaded thorugh service loader)
About this specific module, we need to split better between init (parsing) and apply (runtime). rather than having an Executor that does all the parsing at runtime at the end, you will need several "functor" classes that give you the source, the environmment and so on. 
I will try to comment on specific cases.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunScriptExecutor.java
```diff
@@ -0,0 +1,259 @@
...
+            for (Map.Entry<String, Object> entry :
+                script.getEnvironment().getAdditionalProperties().entrySet()) {
+              String value =
+                  ExpressionUtils.isExpr(entry.getValue())
+                      ? WorkflowUtils.buildStringResolver(
+                              application,
+                              entry.getValue().toString(),
+                              taskContext.input().asJavaObject())
+                          .apply(workflowContext, taskContext, taskContext.input())
+                      : entry.getValue().toString();
```
fjtirado: This is an example of functor I mention in the general comment. 
The idea is that you generate a Map<String,WorkflowValueResolver<String> during init, by using Workfloutils.buildStringFilter (WorkflowApplication appl, String expression), and at runtime you actually invoke the apply (passing the Workflowcontext, TaskContext and WorkflowModel passed to apply method)

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunScriptExecutor.java
```diff
@@ -0,0 +1,259 @@
...
+            source =
+                definition
+                    .resourceLoader()
+                    .load(
+                        scriptUnion.getExternalScript().getSource(),
+                        ResourceLoaderUtils::readString,
+                        workflowContext,
+                        taskContext,
+                        taskContext.input());
+          }
```
fjtirado: This is also another example of "functor" that should be defined at init and later caller at runtime. 
You basically generate a WorkflowValueResolver<String>, so the if are invoked during it, that either give you the source inline or use the workflow and task context to load the resource containing the string

impl/pom.xml
```diff
@@ -92,6 +93,17 @@
                 <artifactId>serverlessworkflow-impl-openapi</artifactId>
                 <version>${project.version}</version>
             </dependency>
+            <dependency>
```
fjtirado, line 98: I think graalvm is a the right choice for javascript library, however, as indicated in the general comment, this dependecy should be in an specific javascript module, not in the core one

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunScriptExecutor.java
```diff
@@ -0,0 +1,259 @@
...
+                case ALL ->
+                    modelFactory.fromAny(
+                        new ProcessResult(
+                            e.getExitStatus(), stdout.toString().trim(), buildStderr(e, stderr)));
+                case NONE -> modelFactory.fromNull();
+                case CODE -> modelFactory.from(e.getExitStatus());
+                case STDOUT -> modelFactory.from(stdout.toString().trim());
+                case STDERR -> modelFactory.from(buildStderr(e, stderr));
+              };
+            }
```
fjtirado: I thing that rather than an external context, GraalVM supports embedded call
Take a look to 
https://www.graalvm.org/latest/reference-manual/js/Modules/#ecmascript-modules-esm

mcruzdev:
> Added a couple of  comments trying to illustrate my previous comment. 

Great comment, I was inclined to create a separate module, especially since the dependency is quite large and really specific. Thanks, I will change it in the next commits.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunScriptExecutor.java
fjtirado: Let me elaborate a bit more on this. The spec states this
_To ensure cross-compatibility, Serverless Workflow strictly limits the versions of supported scripting languages. These versions may evolve with future releases. If you wish to use a different version of a language, you may do so by utilizing the [container process](https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#container-process)._
So this clearly implies that the supported script execution should be embedded. If you want to execute a different version that the supported one, then you use external process. Therefore calling external process to run a supported script is not the way to go.

mathecruz: Hi @fjtirado, about this one, I think it is already embedded. I am investigating a best way to get stderr, I mean, if the script execution throws an exception it will be handled by PolyglotException and we have a lot of builtin methods to get the kind of that exception.

PR #967, opened 2025-11-07
feat: Replace the agent DSL with a composite of instanceId and task name

ricardozanini:

In this PR, we are replacing the `agent` DSL with a unique composite ID referencing the instance ID and the task name, so we guarantee uniqueness to the `@MemoryId` arguments when calling agent methods.

### DIFF
--- a/experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/UniqueIdBiFunction.java
+++ b/experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/UniqueIdBiFunction.java
@@ -15,6 +15,8 @@
  */
 package io.serverlessworkflow.fluent.func.dsl;
 
+import java.util.function.BiFunction;
+
 /**
  * Functions that expect a unique ID injection in runtime, typically an idempotent generated unique
  * id based on the workflow instance id and task name.
@@ -23,6 +25,6 @@ package io.serverlessworkflow.fluent.func.dsl;
  * @param <R> The task result output
  */
 @FunctionalInterface
-public interface UniqueIdBiFunction<T, R> {
+public interface UniqueIdBiFunction<T, R> extends BiFunction<String, T, R> {
   R apply(String uniqueId, T object);
 }


treblereel (approved):
I see no issues with this PR, great job!

experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/FuncDSL.java
```diff
@@ -350,6 +383,31 @@ public static <T, R> FuncCallStep<T, R> withInstanceId(
...
+   * @param in payload input class
+   * @param <T> input type
+   * @param <R> result type
+   * @return a call step
+   */
+  public static <T, R> FuncCallStep<T, R> withUniqueId(
+      String name, UniqueIdBiFunction<T, R> fn, Class<T> in) {
+    JavaFilterFunction<T, R> jff =
+        (payload, wctx, tctx) ->
+            fn.apply(String.format("%s-%s", wctx.instanceData().id(), tctx.taskName()), payload);
```
fjtirado:
```suggestion
            fn.apply(String.format("%s-%s", wctx.instanceData().id(), tctx.position().jsonpointer()), payload);
```

experimental/fluent/func/src/test/java/io/serverlessworkflow/fluent/func/FuncDSLUniqueIdTest.java
```diff
@@ -0,0 +1,260 @@
...
+   */
+  static final class FakeWorkflowContextData implements WorkflowContextData {
+    private final String id;
+
+    FakeWorkflowContextData(String id) {
+      this.id = id;
+    }
+
+    @Override
+    public WorkflowInstanceData instanceData() {
```
fjtirado: I think you should use Mockito rathen than a fake implmenetaiton to make the test shorter

experimental/fluent/func/src/test/java/io/serverlessworkflow/fluent/func/FuncDSLUniqueIdTest.java
```diff
@@ -0,0 +1,260 @@
...
+      return null;
+    }
+
+    @Override
+    public WorkflowDefinitionData definition() {
+      return null;
+    }
+  }
+
+  record FakeTaskContextData(String taskName) implements TaskContextData {
```
fjtirado: Same here, why not use mockito?

experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/UniqueIdBiFunction.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Functions that expect a unique ID injection in runtime, typically an idempotent generated unique
+ * id based on the workflow instance id and task name.
+ *
+ * @param <T> The task payload input
+ * @param <R> The task result output
+ */
+@FunctionalInterface
+public interface UniqueIdBiFunction<T, R> {
```
fjtirado:
```suggestion
public interface UniqueIdBiFunction extends BiFunction<WorkflowContext,TaskContext,String> {
```

And the implementations in the PR should be changed accordingly, We can have a method the generates the default uniqueId, which implementation should be

```
public static String defaultUniqueId (WorkflowContext wctx, TaskContext tctx) {
 return String.format("%s-%s", wctx.instanceData().id(), tctx.position().jsonPointer());
}

```

ricardozanini: I agree about the static method to document the API, but `UniqueIdBiFunction` comes from clients:

```java
  public static <T, R> FuncCallStep<T, R> agent(
      String name, UniqueIdBiFunction<T, R> fn, Class<T> in) {
    return withUniqueId(name, fn, in);
  }
```

Their methods supposed to be like:

```java
function chat(@MemoryId String id, @V String message);
```

There's no internal workflow domain.

fjtirado: Im not completely sure this class is needed, why not just use BiFunction<String,T,R>?. For id search? No matter the reason, this class should extends BiFunction<String,T,R>

fjtirado:
```suggestion
public interface UniqueIdBiFunction<T, R> extends BiFunction<String,T,R> {
```

PR #975, opened 2025-11-14
[Fix #952] Refactoring auth code

### LINKED ISSUE #952: Add  secret support for Oauth2 and OpenID
After https://github.com/serverlessworkflow/sdk-java/issues/928 is completed, it would be good to implement secrets for Oauth2 and OpenId as they were implemented for Basic and Bearer

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/952


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/auth/requestbuilder/AccessTokenProvider.java
```diff
@@ -15,38 +15,46 @@
...
+import jakarta.ws.rs.core.Form;
+import jakarta.ws.rs.core.GenericType;
+import jakarta.ws.rs.core.MediaType;
+import jakarta.ws.rs.core.Response;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.ServiceLoader;
 
 public class AccessTokenProvider {
```
ricardozanini, line 42: Can we replace this provider in the extension? For this feature, we will be using the Quarkus OIDC Client, which we will likely be using.

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/auth/requestbuilder/OAuthRequestBuilder.java
```diff
@@ -15,37 +15,50 @@
...
 import io.serverlessworkflow.api.types.OAuth2AuthenticationPropertiesEndpoints;
 import io.serverlessworkflow.api.types.OAuth2ConnectAuthenticationProperties;
 import io.serverlessworkflow.impl.WorkflowApplication;
+import io.serverlessworkflow.impl.WorkflowUtils;
+import io.serverlessworkflow.impl.WorkflowValueResolver;
 import java.net.URI;
 import java.util.Map;
 
-public class OAuthRequestBuilder extends AbstractAuthRequestBuilder {
+public class OAuthRequestBuilder
```
ricardozanini, line 29: We may even consider this lightweight dep: https://www.pac4j.org/docs/clients/openid-connect.html (maybe as an optional library `impl-http-oidc`)

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/auth/requestbuilder/AccessTokenProvider.java
fjtirado: The Provider suffix here is kind of misleading, but I keep the original because lack of better alternative (Im terrible with naming). This  class is just an aggregation of the info that is needed to perform the call, using JaxRS, to retrieve the token. 
If in future we use Quarkus OIDC client, probably it will be better to write a new HTTP task  (which can be done, base on the quarkus stack, in the quarkus extension)

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/auth/requestbuilder/OAuthRequestBuilder.java
fjtirado: There is not need to add the dependency right now, the call to retrieve the token is based on JaxRS and seems to be working fine.

ricardozanini: But the role OIDC API is complex to implement, have we already done it? cc @treblereel

fjtirado: But its not in the spec schema, isnt it?

ricardozanini: To acquire the token, you need to come back and forth the token server.

The specification includes the scheme to control the endpoints where you'd get the token:
https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#oauth2-authentication

fjtirado: yes, this partially implemented, I will open PR for completion

PR #985, opened 2025-11-17
[Fix #984] Fixing node response type

### LINKED ISSUE #984: HtppExecutor is not deserializing array nodes responses

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/984

### DIFF
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java
@@ -45,9 +45,6 @@ import java.util.function.Supplier;
 
 public class HttpExecutor implements CallableTask<CallHTTP> {
 
-  // TODO allow changing default converter
-  private static final HttpModelConverter defaultConverter = new HttpModelConverter() {};
-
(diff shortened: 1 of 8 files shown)


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java
```diff
@@ -187,16 +187,17 @@ private static RequestSupplier buildRequestSupplier(
...
+          return converter.toModel(
+              application.modelFactory(),
+              node,
+              request.post(converter.toEntity(model), model.objectClass()));
+        };
       case HttpMethod.GET:
       default:
         return (request, w, t, n) ->
-            converter.toModel(application.modelFactory(), n, request.get(n.objectClass()));
+            converter.toModel(application.modelFactory(), n, request.get(Object.class));
```
fjtirado: This should be handled in a more smart way, because although this will work, its doing a lot of Map-JsonNode unneeded conversion. 
Basically converter should be configurable so we can hardcode JsonNode.class there (n.objectClass is not the current response class but the model class, thats the BUG!!!!)

PR #974, opened 2025-11-14
Fix #973 - Refine support for http calls on func

ricardozanini:
**Many thanks for submitting your Pull Request :heart:!**

**What this PR does / why we need it**:
Fix #973

**Special notes for reviewers**:

**Additional information (if needed):**

PR #986, opened 2025-11-17
Fix the HTTP/OpenAPI dsl to include the data transformations

ricardozanini:

Short PR to continue the work from #974

I forgot to add the data transformers to the http/openapi DSLs.


experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/FuncCallOpenAPIStep.java
```diff
@@ -26,65 +26,93 @@
...
+          });
+    } else {
+      list.openapi(
+          name,
+          builder -> {
+            for (Consumer<CallOpenAPITaskFluent<?>> c : steps) {
+              c.accept(builder);
+            }
+            post.accept(builder);
+          });
```
fjtirado: You have basically the same if-else construct in two different places, Im wondering if you can do that logic in the openapi-http method, so basically you check for null there, not here

PR #993, opened 2025-11-19
bug: HttpExecutor was swallowing server URL path; OpenAPIExecutor caching same server URL and operationId

ricardozanini:

In this PR:

- Fix cases where the OpenAPI has a `server.url` with path, e.g. `https://localhost/v1/api`. Paths were returning like `https://localhost/pets` instead of `https://localhost/v1/api/pets`.
- The `OpenAPIExecutor` was caching one `HttpExecutor` per document, not per document + operation. The side effect is that if I have a workflow accessing the same document, but with different operations, only the first operation is called.

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowUtils.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowUtils.java
@@ -30,6 +30,7 @@ import io.serverlessworkflow.impl.resources.ResourceLoader;
 import io.serverlessworkflow.impl.schema.SchemaValidator;
 import io.serverlessworkflow.impl.schema.SchemaValidatorFactory;
 import java.net.URI;
+import java.net.URISyntaxException;
 import java.time.Duration;
 import java.util.Map;
 import java.util.Optional;
@@ -222,7 +223,7 @@ public class WorkflowUtils {
   }
 
   public static URI concatURI(URI base, String pathToAppend) {
-    if (pathToAppend == null || pathToAppend.isEmpty()) {
+    if (!isValid(pathToAppend)) {
       return base;
     }
 
@@ -232,7 +233,7 @@ public class WorkflowUtils {
     }
 
     String basePath = base.getPath();
-    if (basePath == null || basePath.isEmpty()) {
+    if (!isValid(basePath)) {
       basePath = "/";
     } else if (!basePath.endsWith("/")) {
       basePath = basePath + "/";
@@ -253,7 +254,7 @@ public class WorkflowUtils {
 
     try {
       return new URI(base.getScheme(), base.getAuthority(), finalPath, query, fragment);
-    } catch (Exception e) {
+    } catch (URISyntaxException e) {
       throw new IllegalArgumentException(
           "Failed to build combined URI from base=" + base + " and path=" + pathToAppend, e);
     }


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutor.java
```diff
@@ -254,8 +254,41 @@ private static WorkflowValueResolver<WebTarget> getTargetSupplier(
...
+          base.getScheme(),
+          base.getAuthority(),
+          basePath + relPath,
+          path.getQuery(),
+          path.getFragment());
+    } catch (Exception e) {
+      throw new IllegalArgumentException(
+          "Failed to build combined URI from base=" + base + " and path=" + path, e);
+    }
   }
```
fjtirado: Good catch, There is already a WorkflowUtils.concatURI method that should do the trick

```suggestion
return (w, t, n) ->
        HttpClientResolver.client(w, t)
            .target(WorkflowUtils.concatURI(uriSupplier.apply(w, t, n),pathSupplier.apply(w, t, n).toString()));
```

impl/core/src/test/java/io/serverlessworkflow/impl/WorkflowUtilsTest.java
fjtirado, line 1: We can move this test file to core module and test WorkflowUtils.concatURI different use cases

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
```diff
@@ -124,4 +137,65 @@ private void param(String name, Map<String, Object> origMap, Map<String, Object>
...
+    if (!missing.isEmpty()) {
+      String operationId =
+          operation.getOperation().getOperationId() != null
+              ? operation.getOperation().getOperationId()
+              : "<unknown>";
+      throw new IllegalArgumentException(
+          "Missing required OpenAPI parameters for operation '" + operationId + "': " + missing);
+    }
+  }
 }
```
fjtirado, line 201: I do not think this method is really needed, if we want to fail when there is a missing mandatory parameter, we can do in [param](https://github.com/serverlessworkflow/sdk-java/blob/main/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java#L125) method when the param in the target map is null, just adding an else and a throw for the missing params. 
Or, if you want to report all missing params in one shot, add a Map<String,Object> collectMissingParams to that method.

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
```diff
@@ -52,7 +53,10 @@ public void init(CallOpenAPI task, WorkflowDefinition definition) {
     OpenAPIArguments with = task.getWith();
     this.processor = new OpenAPIProcessor(with.getOperationId());
     this.resource = with.getDocument();
-    this.parameters = with.getParameters().getAdditionalProperties();
+    this.parameters =
```
fjtirado, line 56: Paramaters is not mandatory?

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
```diff
@@ -62,20 +66,27 @@ public void init(CallOpenAPI task, WorkflowDefinition definition) {
   @Override
   public CompletableFuture<WorkflowModel> apply(
       WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input) {
+
+    // In the same workflow, access to an already cached document
+    final OperationDefinition operationDefinition =
```
fjtirado, line 71: Ok, I have to admit I cached too much, I think is ok to just cache OperationDefinition and rebuild the HttpExecutor for each call. Its not optimal, but probably is the only choice if paths are modifiable at runtime.

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
ricardozanini: It's not.

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OpenAPIExecutor.java
ricardozanini: Yeah, this is interesting. But in this case, we will miss the case where the parameter is *required*, but has a default (yeah, it happens). Also, going through the list of parameters and showing the list of missing ones is really useful when developing (I experienced that myself while creating the new examples).

fjtirado (approved):

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowUtils.java
```diff
@@ -221,11 +221,42 @@ public static final String checkSecret(
...
+    }
+
+    String finalPath = basePath + relPath;
+
+    String query = child.getQuery();
+    String fragment = child.getFragment();
+
+    try {
+      return new URI(base.getScheme(), base.getAuthority(), finalPath, query, fragment);
+    } catch (Exception e) {
```
fjtirado:
```suggestion
    } catch (URISyntaxException e) {
```

fjtirado (approved):

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowUtils.java
```diff
@@ -221,11 +221,42 @@ public static final String checkSecret(
         .orElseThrow(() -> new IllegalStateException("Secret " + secretName + " does not exist"));
   }
 
-  public static URI concatURI(URI uri, String pathToAppend) {
-    return uri.getPath().endsWith("/")
-        ? uri.resolve(pathToAppend)
-        : URI.create(
-            uri.toString() + (pathToAppend.startsWith("/") ? pathToAppend : "/" + pathToAppend));
+  public static URI concatURI(URI base, String pathToAppend) {
+    if (pathToAppend == null || pathToAppend.isEmpty()) {
```
fjtirado:
```suggestion
    if (!isValid(pathToAppend) {
```

PR #994, opened 2025-11-19
[Fix #836] Implement beforeAgentInvocation and afterAgentInvocation f…

### LINKED ISSUE #836: Implement `beforeAgentInvocation` and `afterAgentInvocation` for LC4J 1.6.0

treblereel:
…or LC4J

https://github.com/serverlessworkflow/sdk-java/issues/836 Implement beforeAgentInvocation and afterAgentInvocation for LC4J 1.6.0

### DIFF
--- a/experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentTaskItemListBuilder.java
+++ b/experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentTaskItemListBuilder.java
@@ -37,9 +37,9 @@ public class AgentTaskItemListBuilder extends BaseTaskItemListBuilder<AgentTaskI
     implements AgentDoFluent<AgentTaskItemListBuilder> {
 
   private final FuncTaskItemListBuilder delegate;
-  private final AtomicReference<Consumer<AgenticScope>> beforeAgentInvocation =
+  private final AtomicReference<Consumer<AgenticScopedRequest>> beforeAgentInvocation =
       new AtomicReference<>();
-  private final AtomicReference<Consumer<AgenticScope>> afterAgentInvocation =
+  private final AtomicReference<Consumer<AgenticScopedResponse>> afterAgentInvocation =
       new AtomicReference<>();
 
   public AgentTaskItemListBuilder() {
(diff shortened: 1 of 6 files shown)


experimental/fluent/agentic/src/main/java/io/serverlessworkflow/fluent/agentic/AgentTaskItemListBuilder.java
```diff
@@ -29,12 +30,17 @@
...
 import java.util.function.Consumer;
 
 public class AgentTaskItemListBuilder extends BaseTaskItemListBuilder<AgentTaskItemListBuilder>
     implements AgentDoFluent<AgentTaskItemListBuilder> {
 
   private final FuncTaskItemListBuilder delegate;
+  private final AtomicReference<Consumer<AgenticScope>> beforeAgentInvocation =
+      new AtomicReference<>();
+  private final AtomicReference<Consumer<AgenticScope>> afterAgentInvocation =
+      new AtomicReference<>();
```
ricardozanini, line 43: Why atomic? Is there a concurrency for changing these builders somewhere?

treblereel:
> Thanks, @treblereel but we also must adapt our implementation:
> 
> https://github.com/serverlessworkflow/sdk-java/blob/main/experimental/fluent/agentic-langchain4j/src/main/java/io/serverlessworkflow/fluent/agentic/langchain4j/ConditionalAgentServiceImpl.java#L53
> 
> It should be part of this PR. Basically, the missing `before|afterAgentInvocation` methods in our LC4J impl.

this one is tough ...

ricardozanini:
> > Thanks, @treblereel but we also must adapt our implementation:
> > https://github.com/serverlessworkflow/sdk-java/blob/main/experimental/fluent/agentic-langchain4j/src/main/java/io/serverlessworkflow/fluent/agentic/langchain4j/ConditionalAgentServiceImpl.java#L53
> > It should be part of this PR. Basically, the missing `before|afterAgentInvocation` methods in our LC4J impl.
> 
> this one is tough ...

I know! That was the real reason why I opened the issue. 😬

Issue #992, opened 2025-11-18
`openapi` functions do not work with JSON files


ricardozanini:
**What happened**:

I tried to create a workflow with an openapi JSON file from Petstore, and it blows:

```
java.util.concurrent.ExecutionException: java.lang.IllegalArgumentException: Invalid OpenAPI document

	at java.base/java.util.concurrent.CompletableFuture.reportGet(CompletableFuture.java:396)
	at java.base/java.util.concurrent.CompletableFuture.get(CompletableFuture.java:2096)
	at org.acme.http.PetstoreFlowIT.petstoreWorkflowShouldReturnPetDetails(PetstoreFlowIT.java:32)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
	at io.quarkus.test.junit.QuarkusTestExtension.runExtensionMethod(QuarkusTestExtension.java:998)
	at io.quarkus.test.junit.QuarkusTestExtension.interceptTestMethod(QuarkusTestExtension.java:846)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
	at java.base/java.util.ArrayList.forEach(ArrayList.java:1511)
Caused by: java.lang.IllegalArgumentException: Invalid OpenAPI document
	at io.serverlessworkflow.impl.executors.openapi.OpenAPIProcessor.getOperation(OpenAPIProcessor.java:43)
	at io.serverlessworkflow.impl.executors.openapi.OpenAPIProcessor.parse(OpenAPIProcessor.java:38)
	at io.serverlessworkflow.impl.executors.openapi.OpenAPIExecutor.lambda$apply$1(OpenAPIExecutor.java:72)
	at io.serverlessworkflow.impl.resources.DefaultResourceLoader.lambda$load$2(DefaultResourceLoader.java:96)
	at java.base/java.util.concurrent.ConcurrentHashMap.compute(ConcurrentHashMap.java:1916)
	at io.serverlessworkflow.impl.resources.DefaultResourceLoader.load(DefaultResourceLoader.java:92)
	at io.serverlessworkflow.impl.executors.openapi.OpenAPIExecutor.apply(OpenAPIExecutor.java:69)
	at io.serverlessworkflow.impl.executors.CallTaskExecutor.internalExecute(CallTaskExecutor.java:58)
	at io.serverlessworkflow.impl.executors.RegularTaskExecutor.execute(RegularTaskExecutor.java:63)
[...]

PR #996, opened 2025-11-20
Add Swagger support to OpenAPI defition processor

treblereel:
fixes https://github.com/serverlessworkflow/sdk-java/issues/992

### DIFF
--- a/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
+++ b/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
@@ -17,26 +17,33 @@ package io.serverlessworkflow.impl.executors.openapi;
 
 import io.swagger.v3.oas.models.OpenAPI;
 import io.swagger.v3.oas.models.Operation;
-import io.swagger.v3.oas.models.media.Content;
 import io.swagger.v3.oas.models.media.MediaType;
 import io.swagger.v3.oas.models.media.Schema;
-import io.swagger.v3.oas.models.parameters.Parameter;
-import io.swagger.v3.oas.models.responses.ApiResponse;
 import io.swagger.v3.oas.models.servers.Server;
+import java.util.ArrayList;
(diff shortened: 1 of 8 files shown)


impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
```diff
@@ -17,26 +17,35 @@
...
 
 class OperationDefinition {
   private final Operation operation;
   private final String method;
   private final OpenAPI openAPI;
   private final String path;
+  private final boolean emulateSwaggerV2BodyParameters;
 
-  OperationDefinition(OpenAPI openAPI, Operation operation, String path, String method) {
+  private static final String APPLICATION_JSON = "application/json";
```
ricardozanini: Since we depend on jax-rs in this module, you can use `MediaType.APPLICATION_JSON` instead.

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/ParameterDefinition.java
```diff
@@ -0,0 +1,50 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.impl.executors.openapi;
+
+import io.swagger.v3.oas.models.parameters.Parameter;
+
+class ParameterDefinition {
```
ricardozanini, line 21: Why can't this be a record? 🤔

treblereel: I expected this question. I didn’t want to make this a record for two reasons: because OperationDefinition is also a class, and because I expect it to have more logic inside, not just be a DTO. And in general, a record is just syntactic sugar around a class :)

PR #1012, opened 2025-11-24
[Fix #1002] Implementing custom inline function call

### LINKED ISSUE #1002: Implement inline function call
Support custom function inline [example](https://github.com/serverlessworkflow/specification/blob/main/examples/call-custom-function-inline.yaml)

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1002

### DIFF
--- a/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaConsumerCallExecutor.java
+++ b/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaConsumerCallExecutor.java
@@ -42,7 +42,7 @@ public class JavaConsumerCallExecutor<T>
     inputClass = task.inputClass();
   }
 
-  public CompletableFuture<WorkflowModel> apply(
+  private CompletableFuture<WorkflowModel> apply(
       WorkflowContext workflowContext, TaskContext taskContext, WorkflowModel input) {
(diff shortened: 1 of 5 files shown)


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
```diff
@@ -0,0 +1,75 @@
...
+  protected WorkflowValueResolver<Map<String, Object>> headersMap;
+  protected WorkflowValueResolver<Map<String, Object>> queryMap;
+  protected Optional<AuthProvider> authProvider = Optional.empty();
+  protected RequestSupplier requestFunction;
+  protected boolean redirect;
+
+  protected static RequestSupplier buildRequestSupplier(
+      String method, Object body, WorkflowApplication application) {
+
+    switch (method.toUpperCase()) {
```
ricardozanini, line 40: Are we supposed to support only POST and GET?

fjtirado: No, we need to complete this. I created https://github.com/serverlessworkflow/sdk-java/issues/1014

PR #1013, opened 2025-11-25
Use redirect property on HTTP and OpenAPI calls

### LINKED ISSUE #959: Add redirect for OpenAPI call
## Context

Today the `OpenAPI` call does not implement the full ServerlessWorkflow specification. The documentation says about the `redirect` property:

> Specifies whether redirection status codes (300–399) should be treated as errors.
If set to false, runtimes must raise an error for response status codes outside the 200–299 range.
If set to true, they must raise an error for status codes outside the 200–399 range.
Defaults to false.

## Tasks

* Add support for `redirect` property on `OpenAPI` call

mcruzdev:
# Changes

* Add a new `RuntimeException` the the `serverlessworkflow-impl-http` module called `RedirectValidationException`.
* Add a new method into `AbstractHttpExecutorBuilder` responsible for applying redirect validation rules for the HTTP response. When the validation fails, it pass the response status code to the `WorkflowError` instance.

## TODO

* Add specific tests for `redirect` behavior (when the upstream server returns a 3XX status code)

Closes #959

### DIFF
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
@@ -44,11 +44,11 @@ abstract class AbstractHttpExecutorBuilder {
         return new WithoutBodyRequestSupplier(Invocation.Builder::delete, application, redirect);
       case HttpMethod.HEAD:
         return new WithoutBodyRequestSupplier(Invocation.Builder::head, application, redirect);
-      case HttpMethod.OPTIONS:
-        return new WithoutBodyRequestSupplier(Invocation.Builder::options, application, redirect);
       case HttpMethod.PATCH:
         return new WithBodyRequestSupplier(
-            (request, entity) -> request.method("patch", entity), application, body, redirect);
+            (request, entity) -> request.method("PATCH", entity), application, body, redirect);
+      case HttpMethod.OPTIONS:
+        return new WithoutBodyRequestSupplier(Invocation.Builder::options, application, redirect);
       case HttpMethod.GET:
       default:
         return new WithoutBodyRequestSupplier(Invocation.Builder::get, application, redirect);
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractRequestSupplier.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractRequestSupplier.java
@@ -15,6 +15,9 @@
  */
 package io.serverlessworkflow.impl.executors.http;
 
+import static jakarta.ws.rs.core.Response.Status.Family.REDIRECTION;
+import static jakarta.ws.rs.core.Response.Status.Family.SUCCESSFUL;
+
 import io.serverlessworkflow.impl.TaskContext;
 import io.serverlessworkflow.impl.WorkflowContext;
 import io.serverlessworkflow.impl.WorkflowError;
@@ -36,6 +39,7 @@ abstract class AbstractRequestSupplier implements RequestSupplier {
   public WorkflowModel apply(
       Builder request, WorkflowContext workflow, TaskContext task, WorkflowModel model) {
     HttpModelConverter converter = HttpConverterResolver.converter(workflow, task);
+
     Response response = invokeRequest(request, converter, workflow, task, model);
     validateStatus(task, response, converter);
     return workflow
@@ -46,7 +50,9 @@ abstract class AbstractRequestSupplier implements RequestSupplier {
   }
 
   private void validateStatus(TaskContext task, Response response, HttpModelConverter converter) {
-    if (response.getStatusInfo().getFamily() != Family.SUCCESSFUL) {
+    Family statusFamily = response.getStatusInfo().getFamily();
+
+    if (statusFamily != SUCCESSFUL || (!this.redirect && statusFamily == REDIRECTION)) {
       throw new WorkflowException(
           converter
               .errorFromResponse(WorkflowError.communication(response.getStatus(), task), response)
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/HTTPWorkflowDefinitionTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/HTTPWorkflowDefinitionTest.java
@@ -18,23 +18,30 @@ package io.serverlessworkflow.impl.test;
 import static io.serverlessworkflow.api.WorkflowReader.readWorkflowFromClasspath;
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.catchThrowableOfType;
+import static org.junit.jupiter.api.Assertions.assertDoesNotThrow;
 
 import io.serverlessworkflow.impl.WorkflowApplication;
 import io.serverlessworkflow.impl.WorkflowModel;
 import java.io.IOException;
+import java.nio.charset.StandardCharsets;
 import java.util.Map;
-import java.util.stream.Stream;
-import org.assertj.core.api.Condition;
+import java.util.concurrent.CompletionException;
+import mockwebserver3.MockResponse;
+import mockwebserver3.MockWebServer;
+import mockwebserver3.RecordedRequest;
+import okhttp3.Headers;
+import org.assertj.core.api.SoftAssertions;
 import org.junit.jupiter.api.AfterAll;
+import org.junit.jupiter.api.AfterEach;
 import org.junit.jupiter.api.BeforeAll;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.Arguments;
-import org.junit.jupiter.params.provider.MethodSource;
-import org.junit.jupiter.params.provider.ValueSource;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Disabled;
+import org.junit.jupiter.api.Test;
 
 public class HTTPWorkflowDefinitionTest {
 
   private static WorkflowApplication appl;
+  private static MockWebServer mockServer;
 
   @BeforeAll
   static void init() {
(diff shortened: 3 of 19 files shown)


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
```diff
@@ -35,27 +39,64 @@ abstract class AbstractHttpExecutorBuilder {
...
+      String method, Object body, boolean redirect, WorkflowApplication application) {
 
     switch (method.toUpperCase()) {
       case HttpMethod.POST:
         WorkflowFilter bodyFilter = WorkflowUtils.buildWorkflowFilter(application, body);
         return (request, w, t, node) -> {
           HttpModelConverter converter = HttpConverterResolver.converter(w, t);
+          Response res =
+              request.buildPost(converter.toEntity(bodyFilter.apply(w, t, node))).invoke();
+
```
mathecruz:
```suggestion
```

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
```diff
@@ -35,27 +39,57 @@ abstract class AbstractHttpExecutorBuilder {
...
+          WorkflowError.communication(
+                  response.getStatus(),
+                  t,
+                  new RedirectValidationException(
+                      String.format(
+                          "The property 'redirect' is %s but received status %d (%s); expected status in the %s range",
+                          redirect,
+                          response.getStatus(),
+                          response.getStatusInfo().getReasonPhrase(),
+                          expectedRange)))
```
fjtirado:
```suggestion
          WorkflowError.communication(
                  response.getStatus(),
                  t,
                      String.format(
                          "The property 'redirect' is %s but received status %d (%s); expected status in the %s range",
                          redirect,
                          response.getStatus(),
                          response.getStatusInfo().getReasonPhrase(),
                          expectedRange))
```

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/RedirectValidationException.java
fjtirado, line 1: I do not think there is need to create a new exception, just the formated message within the error should suffice

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
```diff
@@ -35,27 +39,57 @@ abstract class AbstractHttpExecutorBuilder {
...
+    }
+  }
+
+  private static void applyRedirectValidation(boolean redirect, TaskContext t, Response response) {
+    Response.Status.Family statusFamily = response.getStatusInfo().getFamily();
+    boolean isSuccess = statusFamily.equals(Response.Status.Family.SUCCESSFUL);
+    boolean isRedirect = statusFamily.equals(Response.Status.Family.REDIRECTION);
+
+    boolean valid = redirect ? (isSuccess || isRedirect) : isSuccess;
+    if (!valid) {
```
fjtirado: Lets use import static for Response.Status.Family values

```suggestion
    if (statusFamily !=  SUCCESSFUL || redirect && statusFamily != REDIRECTION) {
```

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
```diff
@@ -46,9 +46,6 @@ protected static RequestSupplier buildRequestSupplier(
         return new WithoutBodyRequestSupplier(Invocation.Builder::head, application, redirect);
       case HttpMethod.OPTIONS:
         return new WithoutBodyRequestSupplier(Invocation.Builder::options, application, redirect);
-      case HttpMethod.PATCH:
-        return new WithBodyRequestSupplier(
-            (request, entity) -> request.method("patch", entity), application, body, redirect);
```
mathecruz, line 51: We need to discuss about `PATCH`, I will write some options tomorrow 🛌🏼

mathecruz: HI folks, I did not have time to solve this one. My new changes:

* Add tests for `DELETE`, `HEAD`, `PUT` and `OPTIONS`
* Use one method by CNCF Specification file
* Remove for now the support for `PATCH`

I removed the support for `PATCH` just to find a better solution for it, actually the HTTP client lib we are using, uses the `java.net.HttpURLConnection` that does not support `PATCH` methods.  To achieve it, we have some workarounds:

* To use a property `HttpUrlConnectorProvider.SET_METHOD_WORKAROUND` but it uses reflection behind the scenes and it is necessary to use  `--add-opens`. (it need an extra configuration)
* Use `X-HTTP-Method-Override` header (in my opnion, it is a bad idea, it is not guaranteed solution)
* Add a new library that supports `PATCH` with no friction.

mathecruz: @ricardozanini @fjtirado do you have another solution in mind?

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractRequestSupplier.java
```diff
@@ -36,20 +39,39 @@ public AbstractRequestSupplier(boolean redirect) {
   public WorkflowModel apply(
       Builder request, WorkflowContext workflow, TaskContext task, WorkflowModel model) {
     HttpModelConverter converter = HttpConverterResolver.converter(workflow, task);
+
+    if (!redirect) {
+      // disable automatic redirects handling from Jersey client
+      request.property("jersey.config.client.followRedirects", false);
```
ricardozanini: We can't tie our implementation to this client. Dowstream client code can override the HTTP client factory, hence it won't be Jersey.

We could have a supplier here where the downstream client can be called to react to this property instead.

By default, we set the jersey property or let the supplier from the underlying client do so.

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
ricardozanini: Jersey won't support it, but RestEasy does, for example. The SDK relies on Jersey by default, but downstream clients can override the `HttpClient`. So it's not always true.

If Jersey, we just fire an `UnsupportedOperation` exception, or let the HTTP client implementation do so. Document it so users know that the default HTTP client shipped with the SDK does not support PATCH.

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractRequestSupplier.java
mathecruz: > We can't tie our implementation to this client. Dowstream client code can override the HTTP client factory, hence it won't be Jersey.

Yeah, I forget this spec/impl detail.

> We could have a supplier here where the downstream client can be called to react to this property instead.

Do you mean a  new method `WorkflowApplication.Builder` method or a  ServiceProvider?

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractRequestSupplier.java
```diff
@@ -36,20 +39,38 @@ public AbstractRequestSupplier(boolean redirect) {
...
-              .errorFromResponse(WorkflowError.communication(response.getStatus(), task), response)
+          WorkflowError.communication(
+                  response.getStatus(),
+                  task,
+                  String.format(
+                      "The property 'redirect' is set to %s but received status %d (%s); expected status in the %s range",
+                      redirect,
+                      response.getStatus(),
+                      response.getStatusInfo().getReasonPhrase(),
+                      expectedRange))
```
fjtirado: Actually, lets keep the prevvious code, 
Why?, because I feel is more interesting to extract the error message from the response (give potentially a lot of more info to the user) rather than explain why we are throwing the error when redirect is enable, that should be self explanatory by looking at the status code

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractRequestSupplier.java
```diff
@@ -36,20 +39,38 @@ public AbstractRequestSupplier(boolean redirect) {
...
 
-  private void validateStatus(TaskContext task, Response response, HttpModelConverter converter) {
-    if (response.getStatusInfo().getFamily() != Family.SUCCESSFUL) {
+  private void validateStatus(TaskContext task, Response response) {
+    Family statusFamily = response.getStatusInfo().getFamily();
+    boolean isSuccess = statusFamily.equals(SUCCESSFUL);
+    boolean isRedirect = statusFamily.equals(REDIRECTION);
+    boolean valid = redirect ? (isSuccess || isRedirect) : isSuccess;
+
+    if (!valid) {
```
fjtirado: I think we can simplify the condition, we want to fail if the call was not successful or if redirect flag is false and the status was a redirection one. 
```suggestion

    if (statusFamily != SUCCESSFUL || !isRedirect && status == REDIRECTION) {
```

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractHttpExecutorBuilder.java
fjtirado: In fact, jersey is only used for testing and explicitly added in examples, it is not really a dependency of the SDK

ricardozanini: @fjtirado but if I use the SDK standalone and once the code reaches `Client.newBuilder`, it explodes if there's no implementation in the classpath?

fjtirado: yes, you need to add a jaxrs compliant library in classpath in order to execute an http request. For test and example, jersey one is added

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/AbstractRequestSupplier.java
fjtirado: We cannot have the client factory in WorkflowApplication.Builder, we need to specify it as additional object. This implies heavy changes [here](https://github.com/serverlessworkflow/sdk-java/blob/main/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpClientResolver.java#L31-L37) and will affect current users. So lets keep it for a different PR (that will also deal with timeouts), in this one you do not need to do anything.

mathecruz: Ok, I will remove the jersey config.

fjtirado (approved):

PR #1019, opened 2025-11-27
[Fix #1018] Initializing context

### LINKED ISSUE #1018: Experimental: null pointer exception when using exportAs
when using this workflow

```
return workflow("investment-memo")
        .tasks(
                get("fetchMarketData", "[http://localhost:8081/market-data/{ticker}](http://localhost:8081/market-data/%7Bticker%7D)")
                        .exportAs(result -> new InvestmentPrompt(
                                result.ticker(), 
                                "",
                                "",
                                ""), MarketDataSnapshot.class)
```

then following exception is thrown

```
2025-11-26 18:23:13,440 ERROR [io.quarkus.vertx.http.runtime.QuarkusErrorHandler] (vert.x-eventloop-thread-1) HTTP Request to /investments/CSU.TO failed, error id: 687f4df4-c30a-49b6-afc6-df9f0c9e664c-1: java.lang.NullPointerException: Cannot invoke "io.serverlessworkflow.impl.WorkflowModel.as(java.lang.Class)" because "n" is null
	at io.serverlessworkflow.impl.expressions.func.JavaExpressionFactory.lambda$buildExpression$1(JavaExpressionFactory.java:47)
	at io.serverlessworkflow.impl.expressions.AbstractExpressionFactory.lambda$buildFilter$5(AbstractExpressionFactory.java:61)
	at io.serverlessworkflow.impl.executors.AbstractTaskExecutor.lambda$apply$11(AbstractTaskExecutor.java:275)
	at java.base/java.util.Optional.ifPresent(Optional.java:178)
	at io.serverlessworkflow.impl.executors.AbstractTaskExecutor.lambda$apply$14(AbstractTaskExecutor.java:272)
	at java.base/java.util.concurrent.CompletableFuture$UniApply.tryFire(CompletableFuture.java:646)
	at java.base/java.util.concurrent.CompletableFuture.postComplete(CompletableFuture.java:510)
	at java.base/java.util.concurrent.CompletableFuture$AsyncSupply.run$$$capture(CompletableFuture.java:1773)
```

When using YAML not initilizing the context is fine because JQ is more tolerant with null than java, but it has to be initialized for java

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1018

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowContext.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowContext.java
@@ -23,6 +23,7 @@ public class WorkflowContext implements WorkflowContextData {
   WorkflowContext(WorkflowDefinition definition, WorkflowMutableInstance instance) {
     this.definition = definition;
     this.instance = instance;
+    this.context = definition.application().modelFactory().fromNull();
   }
 
   @Override
--- a/experimental/lambda/src/test/java/io/serverless/workflow/impl/CallTest.java
+++ b/experimental/lambda/src/test/java/io/serverless/workflow/impl/CallTest.java
@@ -116,7 +116,7 @@ class CallTest {
       Workflow workflow =
           new Workflow()
               .withDocument(
-                  new Document().withNamespace("test").withName("testSwith").withVersion("1.0"))
+                  new Document().withNamespace("test").withName("testSwitch").withVersion("1.0"))
               .withDo(
                   List.of(
                       new TaskItem(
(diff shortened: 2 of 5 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowContext.java
```diff
@@ -23,6 +23,7 @@ public class WorkflowContext implements WorkflowContextData {
   WorkflowContext(WorkflowDefinition definition, WorkflowMutableInstance instance) {
     this.definition = definition;
     this.instance = instance;
+    this.context = definition.application().modelFactory().fromNull();
```
fjtirado, line 26: This is the actual fix (pretty straightforward)
The rest of the PR is adding unit test for context, both for yaml, which was already working without this fix and for Fluent, which was failing without the fix

PR #1021, opened 2025-11-28
Fix #934 Function catalog

### LINKED ISSUE #934: Implement workflow function catalog
As per [spec](https://github.com/serverlessworkflow/specification/blob/main/dsl.md#catalogs)

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/934

### DIFF
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/RetryTimeoutTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/RetryTimeoutTest.java
@@ -136,4 +136,21 @@ public class RetryTimeoutTest {
         .extracting(w -> ((WorkflowException) w.getCause()).getWorkflowError().status())
         .isEqualTo(404);
   }
+
+  @ParameterizedTest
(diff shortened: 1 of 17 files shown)


impl/test/src/test/java/io/serverlessworkflow/impl/test/RetryTimeoutTest.java
```diff
@@ -136,4 +136,21 @@ void testCustomFunction() {
...
+        "workflows-samples/call-custom-function-cataloged-global.yaml"
+      })
+  void testCustomCatalogFunction(String fileName) throws IOException {
+    assertThatThrownBy(
+            () ->
+                app.workflowDefinition(readWorkflowFromClasspath(fileName))
+                    .instance(Map.of())
+                    .start()
+                    .join())
+        .isInstanceOf(IllegalStateException.class)
```
fjtirado, line 153: This way we already have the test when python is implemented

ricardozanini (approved):
Insane!

PR #1027, opened 2025-12-02
[Fix #936] Adding Python implementation

### LINKED ISSUE #936: Implement Python

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/936


impl/python/src/main/java/io/serverlessworkflow/impl/executors/script/python/PythonScriptTaskRunner.java
```diff
@@ -0,0 +1,65 @@
...
+        workflowContext.definition(),
+        ScriptUtils.uncheckedStart(builder),
+        scriptContext.returnType(),
+        model);
+  }
+
+  private void addArg(List<String> list, String name, Object value) {
+    if (value instanceof Boolean bool) {
+      if (bool) {
+        list.add("--" + name);
```
fjtirado, line 58: Because of the way https://github.com/serverlessworkflow/catalog/blob/main/functions/log/1.0.0/function.yaml is implemented, params has to be prefixed with `--`
Notice that this impose restriction in other python scripts to be invoked, which has to follow the same structure. 
Also this is assuming that, for boolean parameters, all python function will use` store_true`
This kind of conventions are one of the reasons why I feel we should change the spec to redefine scripts as calling embedded named parameters functions (which are supported both by JS and Python)

ricardozanini (approved):

impl/pom.xml
```diff
@@ -187,5 +192,6 @@
         <module>container</module>
         <module>test</module>
         <module>script-js</module>
+        <module>python</module>
```
mathecruz, line 195: Why not to follow `script-python` naming to follow an standard?

impl/core/src/main/java/io/serverlessworkflow/impl/scripts/ScriptUtils.java
mathecruz, line 1: Looks great!

impl/python/src/main/java/io/serverlessworkflow/impl/executors/script/python/PythonScriptTaskRunner.java
```diff
@@ -0,0 +1,65 @@
...
+    return ScriptLanguageId.PYTHON;
+  }
+
+  @Override
+  public WorkflowModel runScript(
+      ScriptContext scriptContext,
+      WorkflowContext workflowContext,
+      TaskContext taskContext,
+      WorkflowModel model) {
+    ProcessBuilder builder = new ProcessBuilder("python", "-c", scriptContext.code());
```
mathecruz, line 44: It is just a question, why we are using a process instead a embedded python like we did in Javascript? I mean, what is the motivation?

impl/pom.xml
fjtirado: I think that probably js should have been called javascript

impl/python/src/main/java/io/serverlessworkflow/impl/executors/script/python/PythonScriptTaskRunner.java
fjtirado: It is implied as result of https://github.com/serverlessworkflow/sdk-java/pull/1027#discussion_r2590464004
The catalog function is forcing this to be invoked as external script. 
I do not really like that and I think we need to dicuss on the CNCF forum, but right now we must implement this that way
Also, embedding python will require some intallation procedure in the target machine (the native library to be embedded needs to be compiled by python wheel first), while embedding Javascript is easier because the there are 100% java libraries doing that

PR #1039, opened 2025-12-09
NO-ISSUE: Remove LangChain4j integration

ricardozanini:
The LC4J Agentic API implementation was convoluted - and we pushed to another project where we have infrastructure to support such implementation.

The idea is to focus solely on the engine in the SDK.


experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/FuncForkTaskBuilder.java
```diff
@@ -43,6 +43,7 @@ public class FuncForkTaskBuilder extends TaskBaseBuilder<FuncForkTaskBuilder>
     this.forkTask = new ForkTask();
     this.forkTask.setFork(new ForkTaskConfiguration());
     this.items = new ArrayList<>();
+    this.setTask(forkTask);
```
fjtirado, line 46: This is unrelated, isnt it? Maybe better to fix with another one?

ricardozanini: It's, but I'm lazy to open a new PR because of an one-liner. 👀

PR #1045, opened 2025-12-11
Fixing Swagger 2.1.36 CVE

ricardozanini:
A few libraries from Swagger has CVEs and the upstream project has not upgraded yet. Once they do, we can remove the forced deps.


fjtirado (approved):
I now wondering if we should use swagger to parse the open api file at all (rhino?)

PR #1065, opened 2025-12-19
[Fix #1064] Adding Diggest auth

### LINKED ISSUE #1064: Implement digest authorization
basic, bearer, oauth and openid auth are already implemented, but digest is missing

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1064

### DIFF
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/DigestAuthHttpTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/DigestAuthHttpTest.java
@@ -1,0 +1,136 @@
+/*
+ * Copyright 2020-Present The Serverless Workflow Specification Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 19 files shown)


impl/test/src/test/java/io/serverlessworkflow/impl/test/DigestAuthHttpTest.java
fjtirado, line 1: This test is using values extracted from Wikipedia https://en.wikipedia.org/wiki/Digest_access_authentication#Example_with_explanation

impl/test/src/test/java/io/serverlessworkflow/impl/test/DigestAuthHttpTest.java
```diff
@@ -0,0 +1,136 @@
...
+                      Executors.newFixedThreadPool(
+                          1,
+                          new ThreadFactory() {
+                            @Override
+                            public Thread newThread(Runnable r) {
+                              return new Thread(
+                                  new Runnable() {
+                                    @Override
+                                    public void run() {
+                                      try (MockedStatic<AuthUtils> mocked =
```
fjtirado, line 62: static mocking requires the right thread to be mocked

PR #1076, opened 2025-12-30
Remove io.swagger.parser.v3 to use directly Jackson for deserialization

### LINKED ISSUE #1047: Use a lightweight library for parsing OpenAPI 3 and Swagger v2 to OperationDefinition class
# What happened

Currently, the openapi module are using the library `io.swagger.parser.v3:swagger-parser` that uses a lot of dependencies like `org.mozilla:rhino` (for more info looks the following [comment](https://github.com/serverlessworkflow/sdk-java/pull/1045#issuecomment-3645956377) and the PR)

## Problem

The problem with the current approach is that it increases the final image size from users. To address this issue, we can explore alternative solutions.

# Possible solutions

**Use https://mvnrepository.com/artifact/io.openapiprocessor/openapi-parser**

### Pros

* Minimal dependencies
* Relatively an easy API for use

### Cons

* Does not support Swagger v2, we need to deal with the conversion from Swagger v2 to OpenAPI 3.x.

**Manipulate the OpenAPI 3.x and Swagger v2 manually**

### Pros

* Minimal dependencies (probably 1~3)
* We have the control of the code

### Cons

* Deal with the complexity of parsing the OpenAPI 3.x and Swagger v2 inside the module, actually it is simple because the ServerlessWorkflow Specification only deal with few properties (operation, servers and parameters) but it can turn more complexity in the future with the evolution of the specification.

mcruzdev:
## Changes

This pull request aims to remove io.swagger.parser.v3 dependency to decrease the final image size for sdk-java clients, **these changes only affect the openapi module**.

* Remove io.swagger.parser.v3 dependency
* Add a unified interface for deserialize OpenAPI/Swagger specifications
* Add logic to deserialize a unified OpenAPI/Swagger specification into POJO

Closes #1047

### DIFF
--- a/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/UnifiedOpenAPI.java
+++ b/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/UnifiedOpenAPI.java
@@ -193,9 +193,6 @@ public record UnifiedOpenAPI(
   public record Components(Map<String, Schema> schemas) {}
 
   @JsonIgnoreProperties(ignoreUnknown = true)
-  public record Definitions(Map<String, Schema> definitions) {}
-
-  @JsonIgnoreProperties(ignoreUnknown = true)
   public record Schema(
       String type,
       Map<String, Schema> properties,


fjtirado (approved):

fjtirado:
Looks good, just take a look at the null pointer exception that cause test failure

PR #1089, opened 2026-01-15
Add RequestDecorator to allow SDK clients to modify the request

mcruzdev:
# Changes

Add a `RequestDecorator` interface as SPI to provide a way to configure per-request configuration.

### DIFF
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/RequestDecorator.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/RequestDecorator.java
@@ -15,17 +15,22 @@
  */
 package io.serverlessworkflow.impl.executors.http;
 
+import io.serverlessworkflow.impl.ServicePriority;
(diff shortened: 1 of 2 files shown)


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/RequestDecorator.java
```diff
@@ -0,0 +1,43 @@
...
+import io.serverlessworkflow.impl.WorkflowContext;
+import io.serverlessworkflow.impl.WorkflowModel;
+import jakarta.ws.rs.client.Invocation;
+
+/**
+ * Interface for decorating HTTP requests with additional headers or modifications.
+ *
+ * <p>Implementations should be loaded via ServiceLoader.
+ */
+public interface RequestDecorator {
```
fjtirado:
```suggestion
public interface RequestDecorator extends ServicePriority {
```

This allow user to establish priority between decorators

fjtirado (approved):

PR #1138, opened 2026-02-04
[Fix #1137] Adding Callable Task Proxy builder

### LINKED ISSUE #1137: Allow wrapping of CallableTask
CallableTask contains, as result of building the WorkflowDefintion, the runnable view of a Call Task, which represents a call to an external service
It would be nice to be able to wrap CallableTask execution into existing frameworks, so an application server can run its infrastructure stuff around the remote invocation.

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1137

### DIFF
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/HTTPWorkflowDefinitionTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/HTTPWorkflowDefinitionTest.java
@@ -25,6 +25,7 @@ import io.serverlessworkflow.impl.WorkflowModel;
 import java.io.IOException;
 import java.nio.charset.StandardCharsets;
 import java.util.Map;
+import java.util.concurrent.CompletableFuture;
 import java.util.concurrent.CompletionException;
 import mockwebserver3.MockResponse;
 import mockwebserver3.MockWebServer;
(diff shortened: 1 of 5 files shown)


impl/test/src/test/java/io/serverlessworkflow/impl/test/HTTPWorkflowDefinitionTest.java
```diff
@@ -37,15 +37,29 @@
...
   static void init() {
-    appl = WorkflowApplication.builder().build();
+    appl =
+        WorkflowApplication.builder()
+            .withCallableProxy(
+                delegate ->
+                    (w, t, n) -> {
+                      long init = System.currentTimeMillis();
+                      var result = delegate.apply(w, t, n);
+                      logger.debug(
```
ricardozanini: I'd add a  `logger.isDebugEnabled` to avoid computing time unnecessarily.

fjtirado: formally yes, but this is a test, and probably getting the system clock is faster than checking if debug is enabled anyway ;)

fjtirado: Anyway I changed a bit the test, so calling isDebugEnables really makes a difference

PR #1139, opened 2026-02-05
[Fix #1136] Delegate DB writes to a different thread (keeping write order)

### LINKED ISSUE #1136: Serialize store writes
DB writes are currently performed in the same thread than the task. This is intentional so the next task is not executed till the previous one was written in the db. 
But this approach does not really works well for an implementation with real parallelism, where there can be simoultaneus writes (which in some dbs, can cause locks) 
So, probably it is better to serialize write requests into background thread(s) and let the workflow to continue without waiting for the db to be updated.

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1136


impl/persistence/mvstore/README.md
fjtirado, line 1: These changes are a leftover from previous refactor, same for generared db files

PR #1191, opened 2026-02-27
no-issue: disable tests conditionally when binary is not present

ricardozanini:

Sometimes, a local machine environment might not have `python` or `protoc` installed to run "unit" tests. In my opinion, these tests should be ITs and be on a specific profile. 

However, if we want to run as unit tests and they depend on the environment where they're running, we disable the test if needed.

`mvn clean install` now runs smoothly if python or grpc are not available.

### DIFF
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/PythonCheckCondition.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/PythonCheckCondition.java
@@ -17,16 +17,12 @@ package io.serverlessworkflow.impl.test.junit;
 
 import static io.serverlessworkflow.impl.test.junit.BinaryCheckUtil.isBinaryAvailable;
 
-import java.util.concurrent.ConcurrentHashMap;
-import java.util.concurrent.ConcurrentMap;
 import org.junit.jupiter.api.extension.ConditionEvaluationResult;
 import org.junit.jupiter.api.extension.ExecutionCondition;
 import org.junit.jupiter.api.extension.ExtensionContext;
 import org.junit.platform.commons.util.AnnotationUtils;
 
 public class PythonCheckCondition implements ExecutionCondition {
-
-  private static final ConcurrentMap<String, Boolean> BINARY_CHECKS = new ConcurrentHashMap<>();
 
   @Override
   public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {


impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/BinaryCheckCondition.java
```diff
@@ -0,0 +1,52 @@
...
+
+public class BinaryCheckCondition implements ExecutionCondition {
+
+  @Override
+  public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+    String[] command =
+        context
+            .getElement()
+            .map(el -> el.getAnnotation(DisabledIfBinaryUnavailable.class).value())
+            .orElse(null);
```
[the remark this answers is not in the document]
ricardozanini: This will never happen since Junit won't call this condition when the annotation is not present.

impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/BinaryCheckCondition.java
```diff
@@ -0,0 +1,52 @@
...
+      return ConditionEvaluationResult.disabled(
+          "Test disabled: " + command[0] + " command not found.");
+    }
+  }
+
+  public boolean isBinaryAvailable(String... command) {
+    try {
+      Process process = new ProcessBuilder(command).start();
+      return process.waitFor() == 0;
+    } catch (IOException | InterruptedException e) {
```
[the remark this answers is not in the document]
ricardozanini: Fair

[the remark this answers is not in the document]
ricardozanini: This actually helps.

impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/BinaryCheckUtil.java
```diff
@@ -0,0 +1,53 @@
...
+              return process.exitValue() == 0;
+            }
+            process.destroyForcibly();
+            return false;
+          } catch (IOException | InterruptedException e) {
+            if (e instanceof InterruptedException) {
+              Thread.currentThread().interrupt();
+            }
+            return false;
+          }
```
fjtirado, line 50:
```suggestion
            ProcessBuilder processBuilder =
                new ProcessBuilder(command)
                    .redirectErrorStream(true)
                    .redirectOutput(ProcessBuilder.Redirect.DISCARD);
             Process process = null;
             try {
                    process = process.start();
                    return  process.waitFor(2, TimeUnit.SECONDS) && process.exitValue() == 0;
             } catch (IOException e) {
                  // return default value
             } 
             catch (InterruptedException e) {
                Thread.currentThread().interrupt();
             }
            finally {
                  if (process != null) {
                        process.destroyForcibly();
                  }
           }
          return false;
```

fjtirado: You will need to format this, but the idea is always destroy the process (with current implementation if it is not destroyed if something goes wrong), do not wrap process builder within the finally block and return false except when there is a process to query about. 
And also avoid and unneeded instanceof by using two catch blocks.

fjtirado: Another alternative can be
```suggestion
             Process process;
             try {
                    process = new ProcessBuilder(command)
                    .redirectErrorStream(true)
                    .redirectOutput(ProcessBuilder.Redirect.DISCARD).start();
             } catch (IOException e) {
                  return false;
             } 
             try {
                return  process.waitFor(2, TimeUnit.SECONDS) && process.exitValue() == 0;
             }
             catch (InterruptedException e) {
                Thread.currentThread().interrupt();
                return false;
             }
            finally {
               process.destroyForcibly();
           }
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/ProtocCheckCondition.java
```diff
@@ -0,0 +1,66 @@
...
+        .map(
+            annotation -> {
+              if (isProtocAvailable()) {
+                return ConditionEvaluationResult.enabled(
+                    "Protoc is available for the current platform.");
+              } else {
+                return ConditionEvaluationResult.disabled(
+                    "Test disabled: Protoc not currently available not found.");
+              }
+            })
```
fjtirado:
```suggestion
            annotation -> 
                isProtocAvailable() ? ConditionEvaluationResult.enabled(
                    "Protoc is available for the current platform.") : ConditionEvaluationResult.disabled(
                    "Test disabled: Protoc not currently available not found."))
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/ProtocCheckCondition.java
```diff
@@ -0,0 +1,66 @@
...
+import com.github.os72.protocjar.Protoc;
+import java.util.concurrent.atomic.AtomicReference;
+import org.junit.jupiter.api.extension.ConditionEvaluationResult;
+import org.junit.jupiter.api.extension.ExecutionCondition;
+import org.junit.jupiter.api.extension.ExtensionContext;
+import org.junit.platform.commons.util.AnnotationUtils;
+
+public class ProtocCheckCondition implements ExecutionCondition {
+
+  private static final AtomicReference<Boolean> protocAvailable = new AtomicReference<>(null);
```
fjtirado:
```suggestion
  private static  boolean protocAvailable;
  
  static {
    try {
      protocAvailable = Protoc.runProtoc(new String[] {"--version"}) == 0;
    } catch (Exception e) {
      // ignore
    }
    if (!protocAvailable) {
        protoCAvailable= isBinaryAvailable("protoc", "--version");
     }
  }
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/ProtocCheckCondition.java
```diff
@@ -0,0 +1,66 @@
...
+      if (Protoc.runProtoc(new String[] {"--version"}) == 0) found = true;
+    } catch (Exception e) {
+      // ignore
+    }
+
+    if (!found) found = isBinaryAvailable("protoc", "--version");
+
+    protocAvailable.compareAndSet(null, found);
+
+    return protocAvailable.get();
```
fjtirado:
```suggestion
    return protocAvailable;
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/PythonCheckCondition.java
```diff
@@ -0,0 +1,44 @@
...
+  public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+    return AnnotationUtils.findAnnotation(context.getElement(), DisabledIfPythonUnavailable.class)
+        .map(
+            annotation -> {
+              if (isBinaryAvailable("python", "--version")) {
+                return ConditionEvaluationResult.enabled("Python is available.");
+              } else {
+                return ConditionEvaluationResult.disabled("Test disabled: Python not found.");
+              }
+            })
```
fjtirado:
```suggestion
            annotation -> {
              isBinaryAvailable("python", "--version") ?
                ConditionEvaluationResult.enabled("Python is available.") :
 ConditionEvaluationResult.disabled("Test disabled: Python not found."))
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/junit/ProtocCheckCondition.java
ricardozanini: Must be atomic, it's run by multiple junit forks at the same time.

fjtirado: But being static, it will be loaded at class loading time and it should be thread safe.  Just try, I bet it will  work ;)

PR #1205, opened 2026-03-05
Add number type conversion validation to ensure proper casting between numeric types

### LINKED ISSUE #1200: Incompatible type from task's output and `outputAs`'s input
**What happened**:

Having the following set of tasks:

```java
                            function("scoreProposal", (Proposal input) -> {
                                Integer score = calculateScore(input.abstractText());
                                System.out.println("Score calculated having the result as: " + score);
                                return score;
                            }, Proposal.class)
                                    .outputAs((Integer score) -> {
                                                return new ProposalScore(score, score >= 7);
                                            }, Integer.class),
```

I am receiving the following:

```shell
java.lang.ClassCastException: java.lang.Long incompatible with java.lang.Integer
```

To execute the `outputAs` method.

**How to reproduce it**:

To create a workflow using the provided tasks.

mcruzdev:
# Changes

- Implement `convertNumber` method in AbstractWorkflowModel to handle conversions between numeric types
- Fix issue where asNumber() returned incompatible Number types without proper casting
- Ensure type safety when converting workflow model outputs to specific numeric types

Closes #1200

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/AbstractWorkflowModel.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/AbstractWorkflowModel.java
@@ -24,6 +24,8 @@ public abstract class AbstractWorkflowModel implements WorkflowModel {
 
   protected abstract <T> Optional<T> convert(Class<T> clazz);
 
+  protected abstract <N extends Number> Optional<N> asNumber(Class<N> targetNumberClass);
+
   @Override
   public <T> Optional<T> as(Class<T> clazz) {
     if (WorkflowModel.class.isAssignableFrom(clazz)) {
@@ -35,7 +37,7 @@ public abstract class AbstractWorkflowModel implements WorkflowModel {
     } else if (OffsetDateTime.class.isAssignableFrom(clazz)) {
       return (Optional<T>) asDate();
     } else if (Number.class.isAssignableFrom(clazz)) {
-      return (Optional<T>) asNumber();
+      return (Optional<T>) asNumber(clazz.asSubclass(Number.class));
     } else if (Collection.class.isAssignableFrom(clazz)) {
       Collection<?> collection = asCollection();
       return collection.isEmpty() ? Optional.empty() : (Optional<T>) Optional.of(collection);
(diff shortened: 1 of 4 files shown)


experimental/lambda/src/main/java/io/serverlessworkflow/impl/expressions/func/JavaExpressionFactory.java
```diff
@@ -142,6 +142,28 @@ protected Collection<?> toCollection(Object obj) {
...
+      } else if (clazz == Float.class) {
+        return clazz.cast(num.floatValue());
+      } else if (clazz == Short.class) {
+        return clazz.cast(num.shortValue());
+      } else if (clazz == Byte.class) {
+        return clazz.cast(num.byteValue());
+      }
+    }
     return clazz.cast(obj);
   }
```
fjtirado: hmmm, asClass is a private method that is only invoked for String, CloudEventData, OffsetDataTime, Map and Collection, so this change is not really needed

impl/core/src/main/java/io/serverlessworkflow/impl/AbstractWorkflowModel.java
```diff
@@ -45,4 +45,23 @@ public <T> Optional<T> as(Class<T> clazz) {
...
+      return (T) Double.valueOf(num.doubleValue());
+    } else if (clazz == Float.class) {
+      return (T) Float.valueOf(num.floatValue());
+    } else if (clazz == Short.class) {
+      return (T) Short.valueOf(num.shortValue());
+    } else if (clazz == Byte.class) {
+      return (T) Byte.valueOf(num.byteValue());
+    } else {
+      return (T) num;
+    }
```
fjtirado: I think we need to  handle this differently, rather than using the `asNumber` public interface method, add a new protected method here `asNumber(Class<?> targetNumberClass) `and implement it on the children classes.

That way, you can use `JsonNode.asLong, asInt` methods in  `JakcsonModel `and copy this implementation to` JavaModel`. The approach you took implies two conversions: the one performed by `asNumber` (in JavaModel is straightforward, but in Jackson it is not) and the one performed by this method. It is more efficient to have one in the children class.

impl/core/src/main/java/io/serverlessworkflow/impl/AbstractWorkflowModel.java
```diff
@@ -35,7 +35,7 @@ public <T> Optional<T> as(Class<T> clazz) {
     } else if (OffsetDateTime.class.isAssignableFrom(clazz)) {
       return (Optional<T>) asDate();
     } else if (Number.class.isAssignableFrom(clazz)) {
-      return (Optional<T>) asNumber();
+      return asNumber().map(num -> convertNumber(num, clazz));
```
fjtirado: here is where` asNumber(class) `method should be invoked

Issue #1151, opened 2026-02-10
Add support to subflows on the Java DSL


ricardozanini:
**What would you like to be added**:
Currently, users can't declare workflow calls from the Java DSL. See the language reference here: https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#workflow-process

The runtime engine already supports calling another workflow definition, it's just a matter of supporting it on the Java DSL. Something like:

```java
FuncDSL.workflow().tasks(workflow(myref))
```

**Why is this needed**:
To complete the Java DSL support for the language definition.

PR #1213, opened 2026-03-07
feat: support to subflows on the Java DSL

matheusandre1:

Closes: #1151 

### DIFF
--- a/experimental/lambda/src/test/java/io/serverless/workflow/impl/executors/func/FuncDSLDataFlowTransformationHelpersTest.java
+++ b/experimental/lambda/src/test/java/io/serverless/workflow/impl/executors/func/FuncDSLDataFlowTransformationHelpersTest.java
@@ -163,44 +163,6 @@ public class FuncDSLDataFlowTransformationHelpersTest {
   }
 
   @Test
-  void test_output_with_outputAs() {
-
-    SoftAssertions softly = new SoftAssertions();
-
-    Workflow workflow =
-        FuncWorkflowBuilder.workflow("enrichOutputWithTaskOutputTest")
-            .tasks(
-                function(
-                        "add5",
-                        (Long input) -> {
-                          softly.assertThat(input).isEqualTo(10L);
-                          return input + 5;
-                        },
-                        Long.class)
-                    .outputAs(
-                        (object, workflowContext, taskContextData) -> {
-                          Long taskOutput = output(taskContextData, Long.class);
-                          softly.assertThat(taskOutput).isEqualTo(15L);
-                          Long input = input(workflowContext, Long.class);
-                          softly.assertThat(input).isEqualTo(10L);
-                          return input + taskOutput;
-                        },
-                        Long.class))
-            .build();
-
-    try (WorkflowApplication app = WorkflowApplication.builder().build()) {
-      WorkflowDefinition def = app.workflowDefinition(workflow);
-
-      WorkflowModel model = def.instance(10L).start().join();
-      Number number = model.asNumber().orElseThrow();
-
-      softly.assertThat(number.longValue()).isEqualTo(25L);
-    }
-
-    softly.assertAll();
-  }
-
-  @Test
   void test_input_with_exportAs() {
 
     SoftAssertions softly = new SoftAssertions();
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/SubWorkflowTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/SubWorkflowTest.java
@@ -16,13 +16,14 @@
 package io.serverlessworkflow.impl.test;
 
 import static io.serverlessworkflow.api.WorkflowReader.readWorkflowFromClasspath;
-import static org.junit.Assert.assertEquals;
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.equalTo;
+import static org.hamcrest.Matchers.is;
 
 import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.fluent.spec.WorkflowBuilder;
 import io.serverlessworkflow.fluent.spec.dsl.DSL;
 import io.serverlessworkflow.impl.WorkflowApplication;
-import java.io.IOException;
 import java.util.Map;
 import org.junit.Test;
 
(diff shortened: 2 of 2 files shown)


fluent/spec/src/main/java/io/serverlessworkflow/fluent/spec/dsl/WorkflowSpec.java
```diff
@@ -0,0 +1,71 @@
...
+    steps.add(b -> b.name(name));
+    return this;
+  }
+
+  public WorkflowSpec version(String version) {
+    steps.add(b -> b.version(version));
+    return this;
+  }
+
+  public WorkflowSpec input(Map<String, Object> input) {
```
ricardozanini, line 45: @fjtirado don't we have to support FilterFunction, ContextFunction and Function here too?

fjtirado: That should go on the experimental java side (another PR), this is covering the standard spec part

experimental/lambda/src/test/java/io/serverless/workflow/impl/executors/func/FuncDSLDataFlowTransformationHelpersTest.java
```diff
@@ -160,4 +161,82 @@ void test_input_with_inputFrom_fluent_way() {
...
+
+  @Test
+  void test_output_with_outputAs() {
+
+    SoftAssertions softly = new SoftAssertions();
+
+    Workflow workflow =
+        FuncWorkflowBuilder.workflow("enrichOutputWithTaskOutputTest")
+            .tasks(
+                function(
```
mathecruz: Hi @matheusandre1, it is close. You need to test what you added here. Instead use the `enrichOutputWithTaskTest` workflow use that one you added on `testDoTaskRunWorkflow` test:

```java
    Workflow wf =
        WorkflowBuilder.workflow("parentFlow")
            .tasks(
                d ->
                    d.workflow(
                        "runChild",
                        w ->
                            w.namespace("org.acme")
                                .name("childFlow")
                                .version("1.0.0")
                                .input(Map.of("id", 42, "region", "us-east"))
                                .await(false)
                                .returnType(RunTaskConfiguration.ProcessReturnType.NONE)))
            .build();
```

experimental/lambda/src/test/java/io/serverless/workflow/impl/executors/func/FuncDSLDataFlowTransformationHelpersTest.java
```diff
@@ -160,4 +161,82 @@ void test_input_with_inputFrom_fluent_way() {
...
+                        "add5",
+                        (Long input) -> {
+                          softly.assertThat(input).isEqualTo(10L);
+                          return input + 5;
+                        },
+                        Long.class)
+                    .outputAs(
+                        (object, workflowContext, taskContextData) -> {
+                          Long taskOutput = output(taskContextData, Long.class);
```
fjtirado: actually the object passed to the task is the task ouput, so you should `write (Long taskOutput, worlflowContext, taskContextData) and avoid the call to `output(taskContextData, Long.class);`
@mcruzdev please make sure the usage docs CLEARLY reflec this

experimental/lambda/src/test/java/io/serverless/workflow/impl/executors/func/FuncDSLDataFlowTransformationHelpersTest.java
```diff
@@ -160,4 +161,82 @@ void test_input_with_inputFrom_fluent_way() {
...
+                          softly.assertThat(taskOutput).isEqualTo(15L);
+                          Long input = input(workflowContext, Long.class);
```
fjtirado: This is fine, althoug kind of weird, but I guess make sense in a test

experimental/lambda/src/test/java/io/serverless/workflow/impl/executors/func/FuncDSLDataFlowTransformationHelpersTest.java
mathecruz: The documentation update makes sense, we can do it in another pull request.

I think we need to remove this test, it is duplicated, there is another test that do the same thing in the codebase (@matheusandre1 please remove it).

impl/test/src/test/java/io/serverlessworkflow/impl/test/SubWorkflowTest.java
```diff
@@ -143,4 +145,30 @@ public void outputExportContextAndSetTest() throws IOException {
...
+
+    try (WorkflowApplication app = WorkflowApplication.builder().build()) {
+      app.workflowDefinition(child);
+      Map<String, Object> result =
+          app.workflowDefinition(parent).instance(Map.of()).start().get().asMap().orElseThrow();
+      assertEquals("1", result.get("counter").toString());
+      assertEquals("helloWorld", result.get("greeting").toString());
+    } catch (Exception e) {
+      throw new RuntimeException("Workflow execution failed", e);
+    }
```
fjtirado: You can remove that catch and use join rather than get
In any case, rather than catching and trowing runtime, you can declara that the test function throw any checked exception that might be raised by the code within the function

impl/test/src/test/java/io/serverlessworkflow/impl/test/SubWorkflowTest.java
```diff
@@ -143,4 +145,30 @@ public void outputExportContextAndSetTest() throws IOException {
...
+                    DSL.workflow("org.acme", "childFlow", "1.0.0")
+                        .input(Map.of("id", 42, "region", "us-east"))))
+            .build();
+
+    try (WorkflowApplication app = WorkflowApplication.builder().build()) {
+      app.workflowDefinition(child);
+      Map<String, Object> result =
+          app.workflowDefinition(parent).instance(Map.of()).start().get().asMap().orElseThrow();
+      assertEquals("1", result.get("counter").toString());
+      assertEquals("helloWorld", result.get("greeting").toString());
```
fjtirado: you can change to assertThat() and remove the call toString

impl/test/src/test/java/io/serverlessworkflow/impl/test/SubWorkflowTest.java
matheusandre1: ok, understand, thank you

PR #1289, opened 2026-04-05
fix: Exception on loop after agent call

### LINKED ISSUE #824: Exception on loop after agent call
```
    var creativeWriter = AgentsUtils.newCreativeWriter();
    var scorer = AgentsUtils.newStyleScorer();
    var editor = AgentsUtils.newStyleEditor();

    Predicate<AgenticScope> until = s -> s.readState("score", 0).doubleValue() >= 0.8;

    Map<String, Object> input =
            Map.of(
                    "topic", "fantasy",
                    "style", "fantasy",
                    "audience", "young adults");

    Workflow wf = workflow("loopFlow").agent(creativeWriter)
            .loop(until, scorer, editor).build();

    Object result;
    try (WorkflowApplication app = WorkflowApplication.builder().build()) {
      result = app.workflowDefinition(wf).instance(input).start().get().asJavaObject();
    } catch (Exception e) {
      throw new RuntimeException("Workflow execution failed", e);
    }

    System.out.println(result);
```

**What happened**:
```
Caused by: java.lang.NullPointerException: Cannot invoke "java.util.Optional.isPresent()" because the return value of "io.serverlessworkflow.api.types.func.ForTaskFunction.getForClass()" is null
	at io.serverlessworkflow.impl.executors.func.JavaForExecutorBuilder.collectionFilterObject(JavaForExecutorBuilder.java:70)
	at io.serverlessworkflow.impl.executors.func.JavaForExecutorBuilder.buildCollectionFilter(JavaForExecutorBuilder.java:65)
	at io.serverlessworkflow.impl.executors.ForExecutor$ForExecutorBuilder.<init>(ForExecutor.java:47)
	at io.serverlessworkflow.impl.executors.func.JavaForExecutorBuilder.<init>(JavaForExecutorBuilder.java:37)
	at io.serverlessworkflow.impl.executors.func.JavaTaskExecutorFactory.getTaskExecutor(JavaTaskExecutorFactory.java:30)
	at io.serverlessworkflow.impl.executors.TaskExecutorHelper.createExecutorBuilderList(TaskExecutorHelper.java:93)
	at io.serverlessworkflow.impl.executors.TaskExecutorHelper.createExecutorList(TaskExecutorHelper.java:65)
	at io.serverlessworkflow.impl.WorkflowDefinition.<init>(WorkflowDefinition.java:59)
	at io.serverlessworkflow.impl.WorkflowDefinition.of(WorkflowDefinition.java:69)
	at io.serverlessworkflow.impl.WorkflowDefinition.of(WorkflowDefinition.java:64)
	at io.serverlessworkflow.impl.WorkflowApplication.lambda$workflowDefinition$0(WorkflowApplication.java:257)
	at java.base/java.util.concurrent.ConcurrentHashMap.computeIfAbsent(ConcurrentHashMap.java:1708)
	at io.serverlessworkflow.impl.WorkflowApplication.workflowDefinition(WorkflowApplication.java:256)
	at io.serverlessworkflow.fluent.agentic.langchain4j.WorkflowAgentsIT.loopTest(WorkflowAgentsIT.java:142)
	... 3 more
```

No Exception, resulted String returned from the workflow

matheusandre1:

**What this PR does / why we need it**: Fixes: #824 

**Special notes for reviewers**: modified validations and I set it to optional.empty by default.

### DIFF
--- a/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
+++ b/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
@@ -71,8 +71,7 @@ public class JavaForExecutorBuilder extends ForExecutorBuilder {
   private Object collectionFilterObject(ForTaskFunction taskFunctions) {
     return taskFunctions
         .getForClass()
-        .<Object>map(
-            forClass -> (Object) new TypedFunction(taskFunctions.getCollection(), (Class) forClass))
+        .map(forClass -> new TypedFunction(taskFunctions.getCollection(), forClass))
         .orElse(taskFunctions.getCollection());
   }
 }
--- a/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
+++ b/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
@@ -69,7 +69,7 @@ public class JavaForExecutorBuilder extends ForExecutorBuilder {
   }
 
   private Object collectionFilterObject(ForTaskFunction taskFunctions) {
-    return taskFunctions.getForClass().isPresent()
+    return taskFunctions.getForClass() != null && taskFunctions.getForClass().isPresent()
         ? new TypedFunction(
             taskFunctions.getCollection(), taskFunctions.getForClass().orElseThrow())
         : taskFunctions.getCollection();
--- a/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
+++ b/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
@@ -69,9 +69,10 @@ public class JavaForExecutorBuilder extends ForExecutorBuilder {
   }
 
   private Object collectionFilterObject(ForTaskFunction taskFunctions) {
-    return taskFunctions.getForClass() != null && taskFunctions.getForClass().isPresent()
-        ? new TypedFunction(
-            taskFunctions.getCollection(), taskFunctions.getForClass().orElseThrow())
-        : taskFunctions.getCollection();
+    return taskFunctions
+        .getForClass()
+        .<Object>map(
+            forClass -> (Object) new TypedFunction(taskFunctions.getCollection(), (Class) forClass))
+        .orElse(taskFunctions.getCollection());
   }
 }
--- a/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
+++ b/experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
@@ -71,7 +71,12 @@ public class JavaForExecutorBuilder extends ForExecutorBuilder {
   private Object collectionFilterObject(ForTaskFunction taskFunctions) {
     return taskFunctions
         .getForClass()
-        .map(forClass -> new TypedFunction(taskFunctions.getCollection(), forClass))
+        .<Object>map(forClass -> typedCollectionFunction(taskFunctions, forClass))
         .orElse(taskFunctions.getCollection());
+  }
+
+  @SuppressWarnings({"rawtypes", "unchecked"})
+  private Object typedCollectionFunction(ForTaskFunction taskFunctions, Class<?> forClass) {
+    return new TypedFunction(taskFunctions.getCollection(), forClass);
   }
 }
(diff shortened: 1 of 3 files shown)


experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
```diff
@@ -69,7 +69,7 @@ protected WorkflowValueResolver<Collection<?>> buildCollectionFilter() {
   }
 
   private Object collectionFilterObject(ForTaskFunction taskFunctions) {
-    return taskFunctions.getForClass().isPresent()
+    return taskFunctions.getForClass() != null && taskFunctions.getForClass().isPresent()
```
fjtirado: I think the fix should consist, as you did, in ensuring that getForClass() is always assigned to an Optional. 
This check is defensive programming Im not fan of and I believe we can avoid it

fjtirado (approved):

experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
matheusandre1: Okay, I agree with you, I'll adopt a cleaner version of that.

experimental/lambda/src/main/java/io/serverlessworkflow/impl/executors/func/JavaForExecutorBuilder.java
```diff
@@ -69,9 +69,10 @@ protected WorkflowValueResolver<Collection<?>> buildCollectionFilter() {
...
 
   private Object collectionFilterObject(ForTaskFunction taskFunctions) {
-    return taskFunctions.getForClass().isPresent()
-        ? new TypedFunction(
-            taskFunctions.getCollection(), taskFunctions.getForClass().orElseThrow())
-        : taskFunctions.getCollection();
+    return taskFunctions
+        .getForClass()
+        .<Object>map(
+            forClass -> (Object) new TypedFunction(taskFunctions.getCollection(), (Class) forClass))
```
fjtirado:
```suggestion
            forClass -> new TypedFunction(taskFunctions.getCollection(), (Class) forClass))
```
You can avoid that casting

experimental/types/src/main/java/io/serverlessworkflow/api/types/func/ForTaskFunction.java
```diff
@@ -112,8 +118,8 @@ private <T, V> ForTaskFunction withWhile(
       Optional<Class<?>> modelClass,
       Optional<Class<?>> itemClass) {
     this.whilePredicate = whilePredicate;
-    this.whileClass = modelClass;
-    this.itemClass = itemClass;
+    this.whileClass = modelClass != null ? modelClass : Optional.empty();
+    this.itemClass = itemClass != null ? itemClass : Optional.empty();
```
fjtirado: I do not think modelClass or itemClass can be null here and this method is private. 
This is again defensive programming we should avoid ;)
Note: If this method was public, there will be a chance the parameters were null, but that will be a failure of the caller, and in that case, rather than switching the value provided, you should use Objects.requireNonNull to throw the exception the earlier the better.

experimental/types/src/main/java/io/serverlessworkflow/api/types/func/ForTaskFunction.java
```diff
@@ -147,4 +153,21 @@ public Optional<Class<?>> getItemClass() {
...
+    }
+    if (forClass == null) {
+      forClass = Optional.empty();
+    }
+  }
+
+  private void readObject(ObjectInputStream input) throws IOException, ClassNotFoundException {
+    input.defaultReadObject();
+    normalizeOptionalFields();
+  }
```
fjtirado, line 169: I think this deserve a comment. 
The situation you are trying to prevent is an already serialized definition with null values, because before this PR the fields might not be initialized.

experimental/types/src/main/java/io/serverlessworkflow/api/types/func/ForTaskFunction.java
```diff
@@ -24,11 +26,15 @@ public class ForTaskFunction extends ForTask {
...
-  private Optional<Class<?>> whileClass;
-  private Optional<Class<?>> itemClass;
-  private Optional<Class<?>> forClass;
+  private Optional<Class<?>> whileClass = Optional.empty();
+  private Optional<Class<?>> itemClass = Optional.empty();
+  private Optional<Class<?>> forClass = Optional.empty();
   private Function<?, Collection<?>> collection;
 
+  public ForTaskFunction() {
+    normalizeOptionalFields();
```
fjtirado: You do not need to call this method here (whileClass, itemClass and forClass will be never null because you are now initiliazing them), just for deserialization

fjtirado:
@matheusandre1 can you update the PR with my lastest comment in order to merge it?

fjtirado (approved):

PR #1294, opened 2026-04-08
[NO-ISSUE] Add wait to DSL

mcruzdev:
This pull request introduces a new "wait" task type to the Serverless Workflow Java Fluent API, enabling workflows to pause execution for a specified duration.

**New "wait" task support:**

* Added `WaitTaskBuilder` for configuring wait tasks with duration expressions, inline durations, or timeout builders. (`fluent/spec/src/main/java/io/serverlessworkflow/fluent/spec/WaitTaskBuilder.java`)
* Extended the builder interfaces (`TaskItemListBuilder`, `DoTaskBuilder`, `DoFluent`, and new `WaitFluent`) to support adding wait tasks fluently, including auto-naming and configuration. (`fluent/spec/src/main/java/io/serverlessworkflow/fluent/spec/TaskItemListBuilder.java`, `DoTaskBuilder.java`, `spi/DoFluent.java`, `spi/WaitFluent.java`) [[1]](diffhunk://#diff-79e5a3b6e163ce427671d0d6844a8bb9eb9fa8d4e85ad02373eca62e6b7bf3beR59-R66) [[2]](diffhunk://#diff-8ea722a29ae752a9a3179d142aab9b60660c46d8a655fa6650f34de6d630be3cR82-R87) [[3]](diffhunk://#diff-3aba3316c32d3d508619d7b3dbc1c3f6e8be551d127495889fe3a591f9ee58fcR48) [[4]](diffhunk://#diff-adc4bf7ccf2f07fce002a5502c65425f587da0521fbc631de82e95fce81cf651R1-R28)
* Enhanced the DSL with helper methods for creating wait tasks with various duration specifications. (`fluent/spec/src/main/java/io/serverlessworkflow/fluent/spec/dsl/DSL.java`)

**Testing and validation:**

* Added unit tests to verify wait task auto-naming, duration parsing, and DSL integration. (`TaskItemDefaultNamingTest.java`, `DSLTest.java`) [[1]](diffhunk://#diff-78b008a192c4d9b7dc9485c027b957cd544caba41d48267b48d01cae4840ffa3R132-R148) [[2]](diffhunk://#diff-e4d6a28ecad78b532e0d82f10fa8cf602721931973d7aaacc915e54e8412b3c1R119-R134)
* Introduced integration tests to ensure workflows wait the expected amount of time at runtime. (`impl/test/src/test/java/io/serverlessworkflow/impl/test/WaitTest.java`)

**Bug fix:**

* Fixed a bug in `WaitExecutor` where duration components were not being added correctly, ensuring accurate wait times. (`impl/core/src/main/java/io/serverlessworkflow/impl/executors/WaitExecutor.java`)

**Dependencies:**

* Added the Awaitility library as a test dependency to support timing assertions in wait-related tests. (`impl/test/pom.xml`)


impl/test/pom.xml
```diff
@@ -95,6 +95,11 @@
             <artifactId>grpc-netty</artifactId>
             <scope>test</scope>
         </dependency>
+        <dependency>
+            <groupId>org.awaitility</groupId>
+            <artifactId>awaitility</artifactId>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
fjtirado: It is already there https://github.com/serverlessworkflow/sdk-java/blob/main/impl/test/pom.xml#L89-L92
Basically you need to revert changes on this file

impl/test/src/test/java/io/serverlessworkflow/impl/test/WaitTest.java
fjtirado, line 1: Wait functionality is already covered [here](https://github.com/serverlessworkflow/sdk-java/blob/main/impl/test/src/test/java/io/serverlessworkflow/impl/test/LifeCycleEventsTest.java#L117-L189)

What I think should be done is: 
1) delete this file  (which will susbtantially increase the time running test in case of failure and I made a huge effort to keep the mvn clean install time short)
2) In LifeCycleEventTest class,  to generate, using DSL, a workflow equivalen to wait-set.yaml. 
3) Refactor the code to run the same test with the workflow loaded from the yaml and the workflow generated with the fluent

impl/core/src/main/java/io/serverlessworkflow/impl/executors/WaitExecutor.java
```diff
@@ -45,12 +45,11 @@ protected WaitExecutorBuilder(
...
-      duration.plus(Duration.ofSeconds(durationInline.getSeconds()));
-      duration.plus(Duration.ofMinutes(durationInline.getMinutes()));
-      duration.plus(Duration.ofHours(durationInline.getHours()));
-      duration.plus(Duration.ofDays(durationInline.getDays()));
-      return duration;
+      return Duration.ofMillis(durationInline.getMilliseconds())
+          .plusSeconds(durationInline.getSeconds())
+          .plusMinutes(durationInline.getMinutes())
+          .plusHours(durationInline.getHours())
+          .plusDays(durationInline.getDays());
```
fjtirado, line 52: This is uneeded, but the code looks better, good!!

impl/test/src/test/java/io/serverlessworkflow/impl/test/WaitTest.java
fjtirado: The same approach should be taken for fork-wait.yaml (which will also verify fork dsl is working) 
See https://github.com/serverlessworkflow/sdk-java/blob/main/impl/test/src/test/java/io/serverlessworkflow/impl/test/ForkWaitTest.java#L52-L76

fjtirado: Note that the two test mentioned above, indirectly verify that wait is working without explicit testing, so the test suite still run fast
The target of this PR should be to ensure that we can generate the same type of workflow using fluent that we are generating writing YAML

impl/core/src/main/java/io/serverlessworkflow/impl/executors/WaitExecutor.java
mcruzdev: It was necessary, the `.plus()` does not modify duration, it only creates a new instance of `Duration`, it is immutable.

mcruzdev:
cc: @domhanak

impl/test/src/test/java/io/serverlessworkflow/impl/test/LifeCycleEventsTest.java
```diff
@@ -220,4 +251,31 @@ private <T> T assertPojoInCE(String type, Class<T> clazz) {
...
+
+  private static WorkflowInstance waitSetInstance(WorkflowSource source) throws IOException {
+    return switch (source) {
+      case DSL -> appl.workflowDefinition(waitTestWorkflow()).instance(Map.of());
+      case YAML ->
+          appl.workflowDefinition(
+                  WorkflowReader.readWorkflowFromClasspath("workflows-samples/wait-set.yaml"))
+              .instance(Map.of());
+    };
+  }
```
fjtirado: hmmmm,there should be a different way to do this, the only thing that change here is the Workflow object

impl/test/src/test/java/io/serverlessworkflow/impl/test/ForkWaitTest.java
```diff
@@ -75,11 +78,53 @@ void testForkWaitWithSuspend() throws IOException, InterruptedException {
...
+  private static Stream<Arguments> forkWaitWorkflowSources() {
+    return Stream.of(
+        Arguments.of("yaml", WorkflowSource.YAML), Arguments.of("dsl", WorkflowSource.DSL));
+  }
+
+  private static Workflow forkWaitWorkflow(WorkflowSource source) throws IOException {
+    if (source == WorkflowSource.DSL) {
+      return forkWaitWorkflow();
+    }
+    return readWorkflowFromClasspath("workflows-samples/fork-wait.yaml");
```
fjtirado:
```suggestion
    return source == WorkflowSource.DSL ? forkWaitWorkflow() : readWorkflowFromClasspath("workflows-samples/fork-wait.yaml");
```

fjtirado (approved):

PR #1297, opened 2026-04-10
[Fix #1296] Prevent double loading of the same process instance

### LINKED ISSUE #1296: Prevent loading the same process instance twice
Current persistence API allow loading the same instance twice, which can cause strange side effects on workflow execution. 
Also it would be nice to have a method in WorkflowDefinition to access active instance given an Id, without the use having to deal with Persistence interfaces

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1296

### DIFF
--- a/impl/persistence/tests/src/main/java/io/serverlessworkflow/impl/persistence/test/AbstractHandlerPersistenceTest.java
+++ b/impl/persistence/tests/src/main/java/io/serverlessworkflow/impl/persistence/test/AbstractHandlerPersistenceTest.java
@@ -152,6 +152,7 @@ public abstract class AbstractHandlerPersistenceTest {
     try (Stream<WorkflowInstance> stream = handlers.reader().scanAll(definition)) {
       assertThat(stream.count()).isEqualTo(1);
     }
+    definition.close();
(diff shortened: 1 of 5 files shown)


impl/persistence/tests/src/main/java/io/serverlessworkflow/impl/persistence/test/AbstractHandlerPersistenceTest.java
```diff
@@ -152,6 +152,7 @@ void testWorkflowInstance() throws InterruptedException {
     try (Stream<WorkflowInstance> stream = handlers.reader().scanAll(definition)) {
       assertThat(stream.count()).isEqualTo(1);
     }
+    definition.close();
```
fjtirado, line 155: The fact that I have to do this in order the test to work, it is enough proof that the change is working.
If the instance map is not clear, then the instance is reused and none of the methods are invoked, failing the test.

gmunozfe (approved):
Looks good to me, good catch @fjtirado !

ricardozanini (approved):

PR #1332, opened 2026-04-21
[Fix #1331] withXXX method should call setXXX method

### LINKED ISSUE #1331: Fluent method is not exact replica of setter method
If the generated setXX method is not trivial, his withXXX counterpart wont be aligned. 
For example

```
@OneOfSetter(String.class)
    public void setExpressionErrorDetails(String expressionErrorDetails) {
        this.value = expressionErrorDetails;
        if (expressionErrorDetails_Pattern.matcher(expressionErrorDetails).matches()) {
            this.expressionErrorDetails = expressionErrorDetails;
        } else {
            if (literalErrorDetails_Pattern.matcher(expressionErrorDetails).matches()) {
                this.literalErrorDetails = expressionErrorDetails;
            } else {
                throw new ConstraintViolationException(String.format("%s does not match any pattern", expressionErrorDetails), null);
            }
        }
    }
```

is quite different from

```
public ErrorDetails withExpressionErrorDetails(String value) {
        expressionErrorDetails = value;
        this.value = value;
        return this;
    }
```

To fix that, withXXX should call setXXX

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1331

### DIFF
--- a/fluent/spec/pom.xml
+++ b/fluent/spec/pom.xml
@@ -22,7 +22,7 @@
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
+            <artifactId>junit-jupiter-engine</artifactId>
             <scope>test</scope>
         </dependency>
         <dependency>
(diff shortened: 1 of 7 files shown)


fluent/spec/pom.xml
```diff
@@ -22,7 +22,7 @@
         </dependency>
         <dependency>
             <groupId>org.junit.jupiter</groupId>
-            <artifactId>junit-jupiter-api</artifactId>
+            <artifactId>junit-jupiter-engine</artifactId>
```
fjtirado, line 25: This is to be able to invoke unit test from eclipse

ricardozanini (approved):

mcruzdev (approved):

PR #1334, opened 2026-04-22
[Fix #1333] Fixing WorkflowModelCollection serialization/deserialization

### LINKED ISSUE #1333: Collections models are marshalled as collections and not as model
Which implies that, when unmarshall, they are not really models, but just collections. 
Thats formally uncorrect and might lead to classcastexception when using MarshallingUtils when casting to a WorkflowModel is enforced.

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1333


gmunozfe (approved):
Works like a charm, excellent work @fjtirado !

PR #1340, opened 2026-04-24
[Fix #1339]  Add SerializableFunction support to ForEach on DSL

### LINKED ISSUE #1339: Experimental: Add SerializableFunction support to DSL for each method
Currrently forEach do not have automatic type recgonition, causing classcastexception

See https://github.com/quarkiverse/quarkus-flow/issues/486

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1339

### DIFF
--- a/experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/FuncDSL.java
+++ b/experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/FuncDSL.java
@@ -26,6 +26,7 @@ import io.serverlessworkflow.api.types.FlowDirectiveEnum;
 import io.serverlessworkflow.api.types.OAuth2AuthenticationData;
 import io.serverlessworkflow.api.types.func.ContextFunction;
 import io.serverlessworkflow.api.types.func.FilterFunction;
+import io.serverlessworkflow.api.types.func.LoopFunction;
 import io.serverlessworkflow.fluent.func.FuncCallTaskBuilder;
 import io.serverlessworkflow.fluent.func.FuncEmitTaskBuilder;
 import io.serverlessworkflow.fluent.func.FuncSwitchTaskBuilder;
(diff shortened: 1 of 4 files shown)


experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/FuncDSL.java
```diff
@@ -1064,9 +1020,20 @@ public static FuncTaskConfigurer switchWhenOrElse(
    * @param <T> input type for the collection function
    * @return list configurer
    */
-  public static <T> FuncTaskConfigurer forEach(
-      Function<T, Collection<?>> collection, Consumer<FuncTaskItemListBuilder> body) {
-    return list -> list.forEach(j -> j.collection(collection).tasks(body));
+  public static <T, V> FuncTaskConfigurer forEach(
+      SerializableFunction<T, Collection<V>> collection, Consumer<FuncTaskItemListBuilder> body) {
```
ricardozanini, line 1025: I forgot this? :(

ricardozanini (approved):

experimental/types/src/main/java/io/serverlessworkflow/api/types/func/ForTaskFunction.java
```diff
@@ -29,7 +29,7 @@ public class ForTaskFunction extends ForTask {
   private Optional<Class<?>> whileClass = Optional.empty();
   private Optional<Class<?>> itemClass = Optional.empty();
   private Optional<Class<?>> forClass = Optional.empty();
-  private Function<?, Collection<?>> collection;
+  private Function collection;
```
fjtirado, line 32: I sacrifice compile time type checkign here (which was kind of useless anyway) to have compile time type checking in user code. 
In particular to verify that the type produce by the collection function matches the type expected by the single function parameter loop funcion.

experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/FuncDSL.java
```diff
@@ -1064,9 +1021,25 @@ public static FuncTaskConfigurer switchWhenOrElse(
...
+        list.forEach(
+            j ->
+                j.collection(collection, ReflectionUtils.inferInputType(collection))
+                    .tasks(function));
+  }
+
+  public static <T, V> FuncTaskConfigurer forEachItem(
+      SerializableFunction<T, Collection<V>> collection, Function<V, ?> function) {
+    return forEach(collection, ((t, v) -> function.apply((V) v)));
   }
```
fjtirado, line 1043: Im adding these two ones, which are not directly related with the issue, but I think might be useful to cover most scenarion for users that want to use functions withint the loop (we were already covering the case for other constructs with the existing method)

experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/dsl/FuncDSL.java
fjtirado: Dont be hard on you, I also forgot when reviewing (;

PR #1342, opened 2026-04-24
Add new methods to FuncTryTaskBuilder class

mcruzdev:
This pull request introduces new builder methods to simplify and extend the configuration of try/catch logic in the fluent function workflow API, and adds comprehensive tests to validate these features.

**Enhancements to the try/catch builder API:**

* `FuncTryTaskBuilder` now exposes new public methods for configuring try/catch blocks in a more expressive and flexible way:
  - `try_` for defining the main try tasks
  - `catch_` for specifying error handling logic with custom error selection
  - `catchWhen` for conditional error handling based on expressions
  - `catchType` for handling errors of a specific type

**Testing and validation:**

* Added `FuncTryCatchTest` to thoroughly test the new try/catch builder methods, covering scenarios such as error type matching, conditional error handling, and compensation logic in workflow tasks

### DIFF
--- a/experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/FuncTryTaskBuilder.java
+++ b/experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/FuncTryTaskBuilder.java
@@ -40,7 +40,7 @@ public class FuncTryTaskBuilder
    * @param tryHandler a consumer that configures the tasks to be executed in the try
    * @return this builder instance for method chaining
    */
-  public FuncTryTaskBuilder try_(Consumer<FuncTaskItemListBuilder> tryHandler) {
+  public FuncTryTaskBuilder tryCatch(Consumer<FuncTaskItemListBuilder> tryHandler) {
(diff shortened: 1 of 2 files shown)


experimental/fluent/func/src/main/java/io/serverlessworkflow/fluent/func/FuncTryTaskBuilder.java
```diff
@@ -32,4 +33,54 @@ public class FuncTryTaskBuilder
...
     return this;
   }
+
+  /**
+   * Defines the tasks to execute within the try.
+   *
+   * @param tryHandler a consumer that configures the tasks to be executed in the try
+   * @return this builder instance for method chaining
+   */
+  public FuncTryTaskBuilder try_(Consumer<FuncTaskItemListBuilder> tryHandler) {
```
fjtirado: I think we can use tryTask rather than try_ and catchError rather than catch_, what do you think?

fjtirado (approved):

PR #1350, opened 2026-04-28
Allow create FuncWorkflow with version

### LINKED ISSUE #1349: Allow set workflow version with FuncWorkflowBuilder
It would be great if we have a way to create a `FuncWorkflow` setting version.

mcruzdev:
# Changes

This pull request adds support to set workflow's version when creating a workflow from `FuncWorkflowBuilder`.

Closes #1349


fjtirado:
Im surprised we do not have the version already...

PR #1356, opened 2026-04-30
fix: populate instanceData().output() before onWorkflowCompleted fires

mdproctor:
## Problem

`instanceData().output()` returns `null` when called inside `onWorkflowCompleted` for synchronous workflows.

**Root cause:** In `WorkflowMutableInstance.startExecution()`, the completed `future` is assigned to `futureRef` at line 111 **after** the `CompletableFuture` chain is composed. For synchronous workflows (e.g. simple set-task or Java function tasks), the chain completes inline on the calling thread before `futureRef.set(future)` is reached. `output()` routes through `futureRef.get().join()`, so it returns `null` inside the hook.

Note: `event.output()` on `WorkflowCompletedEvent` already carries the output correctly — that's a separate field passed directly to the constructor. The bug only affects `instanceData().output()`.

## Fix

Store the computed output in a `volatile` field inside `whenSuccess()`, which runs in the chain **before** the completion event is published. `output()` reads that field first, falling back to `futureRef` for callers who retrieve output after `start()` returns.

The change is three lines:

```java
// new field
private volatile WorkflowModel completedOutput;

// in whenSuccess() — store before onWorkflowCompleted fires
this.completedOutput = output;

// in output() — read it first
if (completedOutput != null) return completedOutput;
```

## Test

Adds `WorkflowExecutionListenerOutputTest` with two regression tests:

1. `instanceDataOutputIsAvailableInsideOnWorkflowCompleted` — fails without the fix, passes with it
2. `eventOutputMatchesInstanceDataOutput` — asserts `event.output()` and `instanceData().output()` carry the same value

## Test plan

- [ ] `WorkflowExecutionListenerOutputTest` — both tests pass
- [ ] Full `impl/test` suite — 149 tests, 0 failures

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowMutableInstance.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowMutableInstance.java
@@ -56,6 +56,8 @@ public class WorkflowMutableInstance implements WorkflowInstance {
   private Lock statusLock = new ReentrantLock();
   private Map<CompletableFuture<TaskContext>, TaskContext> suspended;
 
+  private volatile WorkflowModel completedOutput;
+
   private Collection<CompletableFuture<?>> cancelables = new ArrayList<>();
 
   protected WorkflowMutableInstance(WorkflowDefinition definition, String id, WorkflowModel input) {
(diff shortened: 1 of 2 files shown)


mdproctor:
Closing — this was based on a misdiagnosis. `instanceData().output()` is intentionally a blocking join for callers outside the event chain, not the API for use inside event callbacks. The correct API inside `onWorkflowCompleted` is `event.output()`, which is already populated correctly. No fix needed.

PR #1363, opened 2026-05-02
Fix #1354 - Isolate variables in ForExecutor to avoid racing condition to overwrite loop variables

### LINKED ISSUE #1354: ForExecutor race condition: shared TaskContext causes wrong items to be processed in async loops
## Description

The `ForExecutor` has a race condition bug when iterating over collections with >2 items and executing async tasks (like `emit`). The shared `TaskContext` causes variables to be overwritten before async tasks execute, resulting in wrong items being processed.

### Observed Behavior

When using `forEach` to iterate over a collection of 3 items and emit events:

**Input:**
```java
List.of("lint", "test", "build")
```

**Expected:** 3 events emitted for: `lint`, `test`, `build`

**Actual:** 3 events emitted for: `lint`, `build`, `build` (skips "test", duplicates "build")

### Code Analysis

**Location:** `ForExecutor.java` lines 82-90

```java
CompletableFuture<WorkflowModel> future = CompletableFuture.completedFuture(taskContext.input());
while (iter.hasNext()) {
    taskContext.variables().put(task.getFor().getEach(), iter.next());  // Line 83: Sets $item
    taskContext.variables().put(task.getFor().getAt(), i++);            // Line 84: Sets $index
    if (whileExpr.map(w -> w.test(workflow, taskContext, taskContext.input())).orElse(true)) {
        future = future.thenCompose(
            input -> TaskExecutorHelper.processTaskList(
                taskExecutor, workflow, Optional.of(taskContext), input));  // Line 90: Uses SAME taskContext
    }
}
```

**Root Cause:**

1. The `taskContext` object is **shared** across all loop iterations
2. Line 83 sets the loop variable (`$item`) in the shared context
3. Line 90 schedules an **async** task that captures a reference to `taskContext`
4. The while loop continues and **overwrites** the variable with the next item before the async task reads it
5. Result: Async tasks may read the wrong item value (typically the last item in the collection)

### Workflow Definition

```java
import static io.serverlessworkflow.fluent.func.dsl.FuncDSL.*;
import java.util.Collection;
import java.util.List;

public Workflow reproducer() {
    return workflow("forEach-bug-reproducer")
        .tasks(
            // Create a collection of 3 items
            function("createCollection", () -> List.of(
                Map.of("id", "item-1", "name", "first"),
                Map.of("id", "item-2", "name", "second"),
                Map.of("id", "item-3", "name", "third")
            )),
            
            // ForEach should emit 3 events, one per item
            forEach(
                (Collection<Map<String, String>> items) -> items,
                emitJson("test.item.emitted", Map.class)
                    .inputFrom("$item")
            )
        )
        .build();
}
```

### Expected Result

3 CloudEvents emitted with data:
1. `{"id": "item-1", "name": "first"}`
2. `{"id": "item-2", "name": "second"}`
3. `{"id": "item-3", "name": "third"}`

### Actual Result

3 CloudEvents emitted with data:
1. `{"id": "item-1", "name": "first"}` ✓
2. `{"id": "item-3", "name": "third"}` ✗ (should be item-2)
3. `{"id": "item-3", "name": "third"}` ✗ (duplicate)

## Why Existing Tests Don't Catch This

The existing test in `ForTaskFunctionRegressionTest.java` uses synchronous `CallJava` operations:

```java
.withDo(List.of(new TaskItem("javaCall",
    new Task().withCallTask(
        new CallTaskJava(CallJava.loopFunction(CallTest::sum, forConfig.getEach()))))))
```

These complete **before** the loop advances, so the race condition doesn't manifest. The bug only appears with **async** tasks like `emit`, `listen`, or other operations that schedule work via `CompletableFuture`.

## Suggested Fix

The `TaskContext` should be **cloned** or **isolated** per iteration:

```java
while (iter.hasNext()) {
    Object currentItem = iter.next();
    int currentIndex = i++;
    
    // Create isolated context for this iteration
    TaskContext iterationContext = taskContext.copy(); // or new TaskContext(...)
    iterationContext.variables().put(task.getFor().getEach(), currentItem);
    iterationContext.variables().put(task.getFor().getAt(), currentIndex);
    
    if (whileExpr.map(w -> w.test(workflow, iterationContext, taskContext.input())).orElse(true)) {
        future = future.thenCompose(
            input -> TaskExecutorHelper.processTaskList(
                taskExecutor, workflow, Optional.of(iterationContext), input));
    }
}
```

## Impact

This bug affects any workflow using:
- `forEach` with >2 items
- Async tasks within the loop body (emit, listen, etc.)
- Results in data corruption, skipped items, and duplicates

## Environment

- SDK Version: Latest main branch
- Observed in: Quarkus Flow integration with Reactive Messaging (async emit)
- File: `impl/core/src/main/java/io/serverlessworkflow/impl/executors/ForExecutor.java`

ricardozanini:

Fixes #1354 

- I changed the `InMemoryEvents` class slightly to be inherit-friendly
- Moved `TranceExecutionListener` to `impl` so it can easily be reused in other tests and in client code - it's quite useful for debugging/tracing, even in prod code.

### DIFF
--- a/experimental/test/src/test/java/io/serverlessworkflow/fluent/test/ForEachFuncTest.java
+++ b/experimental/test/src/test/java/io/serverlessworkflow/fluent/test/ForEachFuncTest.java
@@ -25,7 +25,7 @@ import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.fluent.func.FuncWorkflowBuilder;
 import io.serverlessworkflow.impl.WorkflowApplication;
 import io.serverlessworkflow.impl.WorkflowModel;
-import io.serverlessworkflow.impl.events.InMemoryEvents;
+import io.serverlessworkflow.impl.lifecycle.TraceExecutionListener;
 import java.time.Duration;
 import java.util.Collection;
 import java.util.List;
@@ -35,13 +35,13 @@ import org.junit.jupiter.api.Test;
 
 public class ForEachFuncTest {
 
-  private static record Order(String id) {}
+  private record Order(String id) {}
 
-  private static record EnhancedOrder(String id, int salary) {}
+  private record EnhancedOrder(String id, int salary) {}
 
-  private static record OrdersPayload(List<Order> orders) {}
+  private record OrdersPayload(List<Order> orders) {}
 
-  private static record OrderName(String id, String name) {}
+  private record OrderName(String id, String name) {}
 
   @Test
   void testForEachIteration() {
@@ -75,13 +75,14 @@ public class ForEachFuncTest {
             .build();
 
     List<CloudEvent> publishedEvents = new CopyOnWriteArrayList<>();
-    InMemoryEvents eventBroker = new InMemoryEvents();
+    LaggedInMemoryEvents eventBroker = new LaggedInMemoryEvents();
     eventBroker.register(eventType, ce -> publishedEvents.add(ce));
 
     try (WorkflowApplication app =
         WorkflowApplication.builder()
             .withEventConsumer(eventBroker)
             .withEventPublisher(eventBroker)
+            .withListener(new TraceExecutionListener())
             .build()) {
       app.workflowDefinition(workflow)
           .instance(
--- a/experimental/test/src/test/java/io/serverlessworkflow/fluent/test/LaggedInMemoryEvents.java
+++ b/experimental/test/src/test/java/io/serverlessworkflow/fluent/test/LaggedInMemoryEvents.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2020-Present The Serverless Workflow Specification Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.fluent.test;
+
+import io.cloudevents.CloudEvent;
+import io.serverlessworkflow.impl.events.InMemoryEvents;
+import java.util.concurrent.CompletableFuture;
+import java.util.function.Consumer;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+public class LaggedInMemoryEvents extends InMemoryEvents {
+
+  private static final Logger logger = LoggerFactory.getLogger(InMemoryEvents.class);
+
+  @Override
+  public CompletableFuture<Void> publish(CloudEvent ce) {
+    return CompletableFuture.runAsync(
+        () -> {
+          Consumer<CloudEvent> allConsumer = allConsumerRef.get();
+          if (allConsumer != null) {
+            allConsumer.accept(ce);
+          }
+          Consumer<CloudEvent> consumer = topicMap.get(ce.getType());
+          if (consumer != null) {
+            consumer.accept(ce);
+          }
+          try {
+            Thread.sleep(10);
+          } catch (InterruptedException e) {
+            throw new RuntimeException(e);
+          }
+          logger.info("Accepted event {} for topic {}", ce.getId(), ce.getType());
+        },
+        serviceFactory.get());
+  }
+}
--- a/experimental/test/src/test/java/io/serverlessworkflow/fluent/test/LaggedInMemoryEvents.java
+++ b/experimental/test/src/test/java/io/serverlessworkflow/fluent/test/LaggedInMemoryEvents.java
@@ -18,33 +18,20 @@ package io.serverlessworkflow.fluent.test;
 import io.cloudevents.CloudEvent;
 import io.serverlessworkflow.impl.events.InMemoryEvents;
 import java.util.concurrent.CompletableFuture;
-import java.util.function.Consumer;
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 
 public class LaggedInMemoryEvents extends InMemoryEvents {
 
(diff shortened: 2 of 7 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/executors/ForExecutor.java
```diff
@@ -80,14 +80,20 @@ protected CompletableFuture<WorkflowModel> internalExecute(
     CompletableFuture<WorkflowModel> future =
         CompletableFuture.completedFuture(taskContext.input());
     while (iter.hasNext()) {
-      taskContext.variables().put(task.getFor().getEach(), iter.next());
-      taskContext.variables().put(task.getFor().getAt(), i++);
+      final Object currentItem = iter.next();
+      final int currentIndex = i++;
+      taskContext.variables().put(task.getFor().getEach(), currentItem);
+      taskContext.variables().put(task.getFor().getAt(), currentIndex);
+
```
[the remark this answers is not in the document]
ricardozanini: This is safe because `thenCompose` chains the futures sequentially - each iteration's future completes only after `processTaskList` and all its subtasks (including parallel branches) finish. The next iteration can't start until the previous one completes, so even though different threads may write to the `HashMap`, they never overlap. The final locals ensure each lambda writes its own captured value right before spawning subtasks, establishing a happens-before relationship that guarantees visibility.

impl/core/src/main/java/io/serverlessworkflow/impl/lifecycle/TraceExecutionListener.java
```diff
@@ -13,19 +13,8 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
-package io.serverlessworkflow.impl.test;
+package io.serverlessworkflow.impl.lifecycle;
```
fjtirado, line 16: The class TraceExecutionListener is only used in a test context, it should not be added to a src directory. This change should be rollback.

fjtirado (commented):
Please put back traceExecutionListerns as test class. 
This will avoid unrelated change in DBGbenrator and MvStorePersistenceTest.

impl/core/src/main/java/io/serverlessworkflow/impl/events/InMemoryEvents.java
```diff
@@ -30,6 +30,10 @@
  */
 public class InMemoryEvents extends AbstractTypeConsumer implements EventPublisher {
 
+  protected final ExecutorServiceFactory serviceFactory;
+  protected final Map<String, Consumer<CloudEvent>> topicMap = new ConcurrentHashMap<>();
+  protected final AtomicReference<Consumer<CloudEvent>> allConsumerRef = new AtomicReference<>();
```
fjtirado: I think this can be done a bit differently
InMemoryEvents can be kept as it is, with the internal state private and then...see my comment in lagged

experimental/test/src/test/java/io/serverlessworkflow/fluent/test/LaggedInMemoryEvents.java
```diff
@@ -0,0 +1,51 @@
...
+          try {
+            Thread.sleep(10);
+          } catch (InterruptedException e) {
+            Thread.currentThread().interrupt();
+            throw new RuntimeException(e);
+          }
+          logger.info("Accepted event {} for topic {}", ce.getId(), ce.getType());
+        },
+        serviceFactory.get());
+  }
```
fjtirado:
```suggestion
    return super.publish(ce).thenRun(() ->  { try {
          Thread.sleep(10);
        } catch (InterruptedException e) {
          Thread.currentThread().interrupt();
        }
        logger.info("Accepted event {} for topic {}", ce.getId(), ce.getType());
      } 
	  );
  }
```

fjtirado: so you do not need to expose inMemoryEvents internal state, just delay the publishing of the event a bit

experimental/test/src/test/java/io/serverlessworkflow/fluent/test/ForEachFuncTest.java
```diff
@@ -35,13 +35,13 @@
...
+  private record Order(String id) {}
 
-  private static record EnhancedOrder(String id, int salary) {}
+  private record EnhancedOrder(String id, int salary) {}
 
-  private static record OrdersPayload(List<Order> orders) {}
+  private record OrdersPayload(List<Order> orders) {}
 
-  private static record OrderName(String id, String name) {}
+  private record OrderName(String id, String name) {}
```
fjtirado, line 44: why not static? I think we should avoid this kind of unneded change, it pollutes the review a lot

experimental/test/src/test/java/io/serverlessworkflow/fluent/test/ForEachFuncTest.java
```diff
@@ -75,13 +75,14 @@ void testForEachEmit() {
...
     List<CloudEvent> publishedEvents = new CopyOnWriteArrayList<>();
-    InMemoryEvents eventBroker = new InMemoryEvents();
+    LaggedInMemoryEvents eventBroker = new LaggedInMemoryEvents();
     eventBroker.register(eventType, ce -> publishedEvents.add(ce));
 
     try (WorkflowApplication app =
         WorkflowApplication.builder()
             .withEventConsumer(eventBroker)
             .withEventPublisher(eventBroker)
+            .withListener(new TraceExecutionListener())
```
fjtirado, line 85: I think the test cover the problem, we should avoid the extra tracing on CI, by removing this

fjtirado:
@ricardozanini Nice pr that fixes that particular race condition, but I realized that ForExecutor was not properly implemented, there are more race conditions and opened this alternate one, please take a look 
https://github.com/serverlessworkflow/sdk-java/pull/1369, if you agree, I think we can close this one

experimental/test/src/test/java/io/serverlessworkflow/fluent/test/ForEachFuncTest.java
ricardozanini: It's static by nature since they are internal classes, we don't need to add the `static` keyword.

fjtirado (approved):

fjtirado (approved):

PR #1370, opened 2026-05-05
[Fix #1351] Generating native file

### LINKED ISSUE #1351: Add native image support
It is a common practise to add a configuration file that indicates which classes will need to be added to a graal-vm native executable to the reflection set. 
See https://www.graalvm.org/22.1/reference-manual/native-image/BuildConfiguration/

fjtirado:
Fix https://github.com/serverlessworkflow/sdk-java/issues/1351


ricardozanini (approved):
Looks good.

PR #1377, opened 2026-05-08
fix: Add until predicates to any listeners

ricardozanini:

- Added tests to validate `toAny` accumulation and predicate evaluation on `until`.
- Enforce specification by throwing an illegal argument exception in case `until` is called in `all` or `one`.


fjtirado (approved):

fjtirado:
@ricardozanini I think the IA might be right and null shoud be allowed for predicate class

PR #1444, opened 2026-06-08
Fix InputBuilder lazy initialization to prevent validation errors

ricardozanini:
## Summary
Fixed `InputBuilder` to use lazy initialization instead of eagerly creating `from` and `schema` objects in the constructor.

## Problem
When users set only a `schema` without `from`, the constructor pre-initialized an empty `InputFrom` object, causing validation errors at runtime:
```
"Both object and str are null" (ObjectExpressionFactory.java:40)
```

Users had to work around this by calling `build().setFrom(null)` explicitly.

## Solution
- Changed constructor to NOT pre-initialize `from` or `schema`
- Made all builder methods null-safe (check before use)
- Ensured proper field clearing when switching between string/object variants
- Added `schemaAsJsonString()` method for JSON string schemas

## Testing
- Added 18 comprehensive unit tests for `InputBuilder`
- All 102 tests in the module pass
- Verified the fix addresses the original issue from the `AgenticFlow` use case

## Changes
- `InputBuilder.java`: Lazy initialization + null-safe methods
- `InputBuilderTest.java`: New comprehensive test suite


ricardozanini:
AI generated the tests and the PR description.

ricardozanini:
@fjtirado why running `mvn clean install` locally changed the db files?

impl/test/db-samples/running.db
fjtirado, line 1: These db files I guess were added uinentionally

ricardozanini: I asked this to you. They were generated with just `mvn clean install`. I think there's something wrong there.

ricardozanini: In the diff, there's only meta changes - the tests are passing.

fjtirado: Yes, I have to take a look, but I think we should not couple it with this PR

ricardozanini: So you will make me copy from main and replace here?

fjtirado: no, just do not include these 4 files in your commit ;)

ricardozanini: Yeah I cant remove it now

PR #1453, opened 2026-06-11
Guard against null parameter fields in OpenAPI operation parsing

mcruzdev:
Guard against null parameter fields in OpenAPI operation parsing                                                                                                                                                         
                                                                                                                                                                                                                           
  Add null checks before dereferencing Parameter.in() and Schema in                                                                                                                                                        
  OperationDefinition, so parameters without an explicit "in" value or                                                                                                                                                     
  schema no longer trigger NullPointerExceptions during parsing.

### DIFF
--- a/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
+++ b/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
@@ -70,7 +70,7 @@ class OperationDefinition {
       }
     }
 
-    if (openAPI.swaggerVersion().equals(UnifiedOpenAPI.SwaggerVersion.SWAGGER_V2)
+    if (openAPI.swaggerVersion() == UnifiedOpenAPI.SwaggerVersion.SWAGGER_V2
         && operation.parameters() != null) {
       operation.parameters().stream()
           .filter(p -> p.in() != null && "body".equals(p.in()))
--- a/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
+++ b/impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
@@ -60,7 +60,7 @@ class OperationDefinition {
     List<ParameterDefinition> paramDefinitions = new ArrayList<>();
     if (operation.hasParameters()) {
       for (UnifiedOpenAPI.Parameter parameter : operation.parameters()) {
-        if (parameter.in() != null && "body".equals(parameter.in())) {
+        if ("body".equals(parameter.in())) {
           continue; // body parameters are handled separately
         }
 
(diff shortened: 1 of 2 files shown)


fjtirado (approved):
@mcruzdev Good.
Lets fix AI comments

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
```diff
@@ -60,7 +60,7 @@ List<ParameterDefinition> getParameters() {
     List<ParameterDefinition> paramDefinitions = new ArrayList<>();
     if (operation.hasParameters()) {
       for (UnifiedOpenAPI.Parameter parameter : operation.parameters()) {
-        if (parameter.in().equals("body")) {
+        if (parameter.in() != null && "body".equals(parameter.in())) {
```
fjtirado:
```suggestion
        if ("body".equals(parameter.in())) {
```

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
```diff
@@ -70,13 +70,14 @@ List<ParameterDefinition> getParameters() {
       }
     }
 
-    if (openAPI.swaggerVersion().equals(UnifiedOpenAPI.SwaggerVersion.SWAGGER_V2)) {
+    if (openAPI.swaggerVersion() == UnifiedOpenAPI.SwaggerVersion.SWAGGER_V2
+        && operation.parameters() != null) {
       operation.parameters().stream()
-          .filter(p -> p.in().equals("body"))
+          .filter(p -> p.in() != null && "body".equals(p.in()))
```
fjtirado:
```suggestion
          .filter(p -> "body".equals(p.in()))
```

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/UnifiedOpenAPI.java
```diff
@@ -165,7 +165,11 @@ public boolean hasRequestBody() {
     }
   }
 
-  public record Parameter(String name, String in, Boolean required, Schema schema) {}
+  public record Parameter(String name, String in, Boolean required, Schema schema) {
```
fjtirado: Im wondering if we can make required field a boolean rather than a Boolean and skip that check

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/UnifiedOpenAPI.java
```diff
@@ -165,7 +165,11 @@ public boolean hasRequestBody() {
...
+    public Parameter {
+      required = required != null ? required : Boolean.FALSE;
+    }
+  }
```
fjtirado: Yes, I tested it and it is working 

```suggestion
  public record Parameter(String name, String in, boolean required, Schema schema) {
  }
```

impl/openapi/src/main/java/io/serverlessworkflow/impl/executors/openapi/OperationDefinition.java
```diff
@@ -70,13 +70,14 @@ List<ParameterDefinition> getParameters() {
       }
     }
 
-    if (openAPI.swaggerVersion().equals(UnifiedOpenAPI.SwaggerVersion.SWAGGER_V2)) {
+    if (openAPI.swaggerVersion() == UnifiedOpenAPI.SwaggerVersion.SWAGGER_V2
+        && operation.parameters() != null) {
```
fjtirado:
```suggestion
        && operation.hasParameters()) {
```

PR #1473, opened 2026-06-21
Refactor catch task executor creation to include position property

### LINKED ISSUE #1472: Try task generates wrong jsonPosition for the nested catch task
While executing the attached process `try-catch-with-retry-and-do`  the following behaviour was observed.

The tryTask gets the following position: do/0/tryTask/try which is correct

The faillngTask gets the follwoing jsonPositon:  do/0/tryTask/try/0/failingTask  (see that it's preceeded with the parent position  do/0/tryTask/try) which is correct.

However, the executeAfterFailingTask gets the following jsonPosition:   do/0/tryTask/catch/do/0/executeAfterFailingTask . 

See that there's no way to relate it with the parent.

Since the task is derived from the tryTask execution, it must get a jsonPositon like this:   do/0/tryTask/try/catch/do/0/executeAfterFailingTask 

[test-try-catch-wity-retry-and-do.yaml](https://github.com/user-attachments/files/29159160/test-try-catch-wity-retry-and-do.yaml)

matheusandre1:
Fixes: #1472


fjtirado:
See comment on issue, closing this for now.

PR #1487, opened 2026-06-25
Allow injecting a custom AuthProviderFactory to let frameworks own authentication

### LINKED ISSUE #1486: Allow injecting a custom AuthProviderFactory to let frameworks own authenticatio
# Context

Today the SDK owns authentication: when a task has an auth policy, it builds the AuthProvider and performs the OAuth2/OIDC token exchange itself.                 
                                                                   
When running inside a framework (Quarkus) that already manages identity, the framework should be able to take over and produce the Authorization value, suppressing the SDK's own token exchange.

# Proposal                           

Make AuthProviderFactory pluggable so a custom implementation can be provided, with the current behavior as the default when none custom `AuthProviderFactory` is registered.

mcruzdev:

This pull request allows a framework to inject a own AuthProviderFactory to control how the exchange token is made.

Closes #1486

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
@@ -22,6 +22,8 @@ import io.serverlessworkflow.api.types.SchemaInline;
 import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.impl.additional.NamedWorkflowAdditionalObject;
 import io.serverlessworkflow.impl.additional.WorkflowAdditionalObject;
+import io.serverlessworkflow.impl.auth.AuthProviderFactory;
+import io.serverlessworkflow.impl.auth.DefaultAuthProviderFactory;
 import io.serverlessworkflow.impl.config.ConfigManager;
 import io.serverlessworkflow.impl.config.ConfigSecretManager;
 import io.serverlessworkflow.impl.config.SecretManager;
@@ -89,6 +91,7 @@ public class WorkflowApplication implements AutoCloseable {
   private final WorkflowModelFactory contextFactory;
   private final WorkflowScheduler scheduler;
   private final Map<String, WorkflowAdditionalObject<?>> additionalObjects;
+  private final AuthProviderFactory authProviderFactory;
   private final ConfigManager configManager;
   private final SecretManager secretManager;
   private final SchedulerListener schedulerListener;
@@ -120,6 +123,7 @@ public class WorkflowApplication implements AutoCloseable {
     this.scheduler = builder.scheduler;
     this.schedulerListener = builder.schedulerListener;
     this.additionalObjects = builder.additionalObjects;
+    this.authProviderFactory = builder.authProviderFactory;
     this.configManager = builder.configManager;
     this.secretManager = builder.secretManager;
     this.templateResolver = builder.templateResolver;
@@ -241,6 +245,7 @@ public class WorkflowApplication implements AutoCloseable {
     private WorkflowModelFactory modelFactory;
     private WorkflowModelFactory contextFactory;
     private Map<String, WorkflowAdditionalObject<?>> additionalObjects = new HashMap<>();
+    private AuthProviderFactory authProviderFactory = DefaultAuthProviderFactory.INSTANCE;
     private SecretManager secretManager;
     private ConfigManager configManager;
     private SchedulerListener schedulerListener;
@@ -354,6 +359,11 @@ public class WorkflowApplication implements AutoCloseable {
     public <T> Builder withAdditionalObject(
         String name, WorkflowAdditionalObject<T> additionalObject) {
       additionalObjects.put(name, additionalObject);
+      return this;
+    }
+
+    public Builder withAuthProviderFactory(AuthProviderFactory authProviderFactory) {
+      this.authProviderFactory = authProviderFactory;
       return this;
     }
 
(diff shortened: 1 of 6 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -241,6 +245,7 @@ public SchemaValidator getValidator(SchemaInline inline) {
     private WorkflowModelFactory modelFactory;
     private WorkflowModelFactory contextFactory;
     private Map<String, WorkflowAdditionalObject<?>> additionalObjects = new HashMap<>();
+    private AuthProviderFactory authProviderFactory = DefaultAuthProviderFactory.INSTANCE;
```
fjtirado: Lets follow the same pattern than other factories, do not initialize the variable in the builder and check later for null and then, if null, use the default instance. 
Also, for default instance, lets use the double lazy initialization pattern, like here 
https://github.com/serverlessworkflow/sdk-java/blob/main/impl/core/src/main/java/io/serverlessworkflow/impl/marshaller/DefaultBufferFactory.java#L36-L38
That way the instance is only initialized if it is really used. 
Alternative you can completely avoid the singleton pattern and just create an instance as it is done for other factories https://github.com/serverlessworkflow/sdk-java/blob/main/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java#L457-L459

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutorBuilder.java
```diff
@@ -101,24 +101,13 @@ public static HttpExecutorBuilder builder(WorkflowDefinition definition) {
...
-      case HttpMethod.HEAD:
-      case HttpMethod.OPTIONS:
-      case HttpMethod.GET:
-      default:
-        return new WithoutBodyRequestExecutor(
-            theMethod, redirect, AuthProviderFactory.getAuth(definition, policy, method));
-    }
+    String httpMethod = method.toUpperCase();
+    Optional<AuthProvider> auth =
+        definition.application().authProviderFactory().getAuth(definition, policy, method);
```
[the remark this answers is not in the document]
fjtirado: yes, AI is right

impl/test/src/test/java/io/serverlessworkflow/impl/test/CustomAuthProviderFactoryOverrideTest.java
fjtirado, line 1: I do not like this test, too verbose and does not really add anythign since the code is pretty trivial, but since it is already written...

impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
```diff
@@ -0,0 +1,103 @@
...
+import io.serverlessworkflow.api.types.Use;
+import io.serverlessworkflow.api.types.UseAuthentications;
+import io.serverlessworkflow.api.types.Workflow;
+import io.serverlessworkflow.impl.WorkflowApplication;
+import io.serverlessworkflow.impl.WorkflowDefinition;
+import java.util.Optional;
+
+public class DefaultAuthProviderFactory implements AuthProviderFactory {
+
+  public static final DefaultAuthProviderFactory INSTANCE = new DefaultAuthProviderFactory();
```
fjtirado: On a second thought, I think we should about the singleton pattern for this class, since it is very likely that a smart framework delevoper might reuse it by using inheritance and the singleton kind of ruins that.

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -357,6 +362,11 @@ public <T> Builder withAdditionalObject(
       return this;
     }
 
+    public Builder withAuthProviderFactory(AuthProviderFactory authProviderFactory) {
+      this.authProviderFactory = authProviderFactory;
+      return this;
+    }
```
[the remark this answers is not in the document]
fjtirado: This is addresed by doing what I propose here https://github.com/serverlessworkflow/sdk-java/pull/1487/changes#r3473020618

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutorBuilder.java
```diff
@@ -101,24 +101,13 @@ public static HttpExecutorBuilder builder(WorkflowDefinition definition) {
...
-        return new WithoutBodyRequestExecutor(
-            theMethod, redirect, AuthProviderFactory.getAuth(definition, policy, method));
-    }
+    String httpMethod = method.toUpperCase();
+    Optional<AuthProvider> auth =
+        definition.application().authProviderFactory().getAuth(definition, policy, httpMethod);
+    return switch (httpMethod) {
+      case HttpMethod.POST, HttpMethod.PUT, HttpMethod.PATCH ->
+          new WithBodyRequestExecutor(httpMethod, redirect, auth, definition.application(), body);
+      default -> new WithoutBodyRequestExecutor(httpMethod, redirect, auth);
```
fjtirado: This is a nitpick, but I liked the list of operations that do not have body to be explicitly stated, somehow it was easier to understand

fjtirado (approved):

impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutorBuilder.java
mcruzdev: I am with you

fjtirado (approved):

PR #1491, opened 2026-06-26
Validate task's auth being referencing a not defined use.authentications

### LINKED ISSUE #1490: Referencing an undefined authentication via `.use` silently sends request unauthenticated
When a task references an authentication by name (e.g. `authentication.use: sampleDigest`) and that name is not defined in the `use.authentications` block, the request is silently sent without
   authentication instead of failing.

   The fix should validate at build time that the referenced authentication name exists in `use.authentications` and throw an error if it doesn't.

mcruzdev:

Closes #1490

### DIFF
--- a/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutorBuilder.java
+++ b/impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/HttpExecutorBuilder.java
@@ -16,6 +16,8 @@
 package io.serverlessworkflow.impl.executors.http;
 
 import io.serverlessworkflow.api.types.ReferenceableAuthenticationPolicy;
+import io.serverlessworkflow.api.types.Use;
+import io.serverlessworkflow.api.types.UseAuthentications;
 import io.serverlessworkflow.impl.WorkflowDefinition;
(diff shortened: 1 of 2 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
```diff
@@ -65,16 +65,15 @@ public Optional<AuthProvider> getAuth(
   private Optional<AuthProvider> buildFromReference(
       WorkflowApplication app, Workflow workflow, String use, String method) {
     Use useInfo = workflow.getUse();
-    if (useInfo == null) {
-      return Optional.empty();
+    UseAuthentications authInfo = useInfo == null ? null : useInfo.getAuthentications();
+    AuthenticationPolicyUnion policy =
+        authInfo == null ? null : authInfo.getAdditionalProperties().get(use);
```
fjtirado: I like this approach to have a compact code, but you can have left the current implementation of this method untouched, and just add an ifEmpty check to the caller and throw the exception there. 
Lets keep it this way (there is no action associated to this comment), because the check is previous to the creation of the optional (the sooner you fail the better)

fjtirado (approved):

PR #1512, opened 2026-07-01
Expose auth policy resolution utilities for framework integrations

### LINKED ISSUE #1511: Expose reference resolution in DefaultAuthProviderFactory as a public static method
## Summary

   `DefaultAuthProviderFactory` bundles several concerns that framework integrations need individually: (1) resolving `ReferenceableAuthenticationPolicy` references, (2) extracting
   `OAuth2AuthenticationData` from the union-of-unions structure, and (3) constructing `AuthProvider` instances. All internal methods are private, forcing frameworks to duplicate the resolution and
   extraction logic.

   ## Deliverables

   ### 1. `resolvePolicy` static method

   Public static method on `DefaultAuthProviderFactory` that resolves a `ReferenceableAuthenticationPolicy` to an `AuthenticationPolicyUnion`:

   ```java
   public static AuthenticationPolicyUnion resolvePolicy(
       Workflow workflow, ReferenceableAuthenticationPolicy auth)
   ```

   ### 2. `OAuthPolicyData` record

   Public record that extracts `OAuth2AuthenticationData` + `SecretBasedAuthenticationPolicy` + OIDC flag from an `AuthenticationPolicyUnion`, eliminating the need to navigate the union-of-unions structure:

   ```java
   public record OAuthPolicyData(
       OAuth2AuthenticationData data,
       SecretBasedAuthenticationPolicy secret,
       boolean openIdConnect) {
     public static Optional<OAuthPolicyData> from(AuthenticationPolicyUnion policy);
   }
   ```

   Both changes are:
   - **Additive-only** — no existing API changes
   - **Fully backward compatible** — new public types/methods, no signature changes
   - **Pure and stateless** — no dependency on instance state

   The existing `buildFromPolicy` is refactored internally to use `OAuthPolicyData.from()`, consolidating the OAuth2/OIDC dispatch.

   ## Use case

   The [Quarkus Flow](https://github.com/quarkiverse/quarkus-flow) OIDC extension implements `AuthProviderFactory` using Quarkus OIDC Client for token negotiation. It currently maintains a local
   `OAuth2Policy` type that duplicates the union unwrapping and a `union()` method that duplicates the reference resolution logic. With `resolvePolicy()` + `OAuthPolicyData.from()`, the framework
   integration pipeline becomes:

   ```java
   AuthenticationPolicyUnion union = DefaultAuthProviderFactory.resolvePolicy(workflow, auth);
   Optional<OAuthPolicyData> oauthData = OAuthPolicyData.from(union);
   // oauthData.get().data() → OAuth2AuthenticationData
   // oauthData.get().openIdConnect() → boolean
   ```

   This eliminates the Quarkus-side `OAuth2Policy` type and the `union()` reference resolution method entirely.

mcruzdev:

Closes #1511

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
@@ -54,6 +54,10 @@ public class DefaultAuthProviderFactory implements AuthProviderFactory {
 
   public static AuthenticationPolicyUnion resolvePolicy(
       Workflow workflow, ReferenceableAuthenticationPolicy auth) {
+    if (workflow == null) {
+      throw new IllegalArgumentException(
+          "workflow must not be null when resolving an authentication policy reference");
+    }
     if (auth == null) {
       return null;
     }
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
@@ -86,16 +86,12 @@ public class DefaultAuthProviderFactory implements AuthProviderFactory {
       return Optional.of(
           new DigestAuthProvider(
               app, workflow, authenticationPolicy.getDigestAuthenticationPolicy(), method));
-    } else if (authenticationPolicy.getOAuth2AuthenticationPolicy() != null) {
-      return Optional.of(
-          new OAuth2AuthProvider(
-              app, workflow, authenticationPolicy.getOAuth2AuthenticationPolicy()));
-    } else if (authenticationPolicy.getOpenIdConnectAuthenticationPolicy() != null) {
-      return Optional.of(
-          new OpenIdAuthProvider(
-              app, workflow, authenticationPolicy.getOpenIdConnectAuthenticationPolicy()));
     }
-
-    return Optional.empty();
+    return OAuthPolicyData.from(authenticationPolicy)
+        .map(
+            policyData ->
+                policyData.scheme() == OAuthPolicyData.OAuthScheme.OPENID_CONNECT
+                    ? new OpenIdAuthProvider(app, workflow, policyData)
+                    : new OAuth2AuthProvider(app, workflow, policyData));
   }
 }
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
@@ -18,8 +18,6 @@ package io.serverlessworkflow.impl.auth;
 import io.serverlessworkflow.api.types.AuthenticationPolicyUnion;
 import io.serverlessworkflow.api.types.EndpointConfiguration;
 import io.serverlessworkflow.api.types.ReferenceableAuthenticationPolicy;
-import io.serverlessworkflow.api.types.Use;
-import io.serverlessworkflow.api.types.UseAuthentications;
 import io.serverlessworkflow.api.types.Workflow;
 import io.serverlessworkflow.impl.WorkflowApplication;
 import io.serverlessworkflow.impl.WorkflowDefinition;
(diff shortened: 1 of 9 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
```diff
@@ -46,35 +46,31 @@ public Optional<AuthProvider> getAuth(
...
+        ? Optional.empty()
+        : buildFromPolicy(definition.application(), definition.workflow(), policy, method);
+  }
+
+  public static AuthenticationPolicyUnion resolvePolicy(
+      Workflow workflow, ReferenceableAuthenticationPolicy auth) {
+    if (workflow == null) {
+      throw new IllegalArgumentException(
+          "workflow must not be null when resolving an authentication policy reference");
+    }
```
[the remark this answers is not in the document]
fjtirado: Ill add that is is absurd for workflow to be null here

impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
```diff
@@ -46,35 +46,27 @@ public Optional<AuthProvider> getAuth(
...
-  private Optional<AuthProvider> buildFromReference(
-      WorkflowApplication app, Workflow workflow, String use, String method) {
-    Use useInfo = workflow.getUse();
-    if (useInfo == null) {
-      return Optional.empty();
+      String use = auth.getAuthenticationPolicyReference().getUse();
+      Use useInfo = workflow.getUse();
+      if (useInfo == null) {
+        return null;
+      }
```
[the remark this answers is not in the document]
fjtirado: I do not agree with that one. AI seems to love stupid defensive programing, Workflow cannot be null, use can be null

impl/core/src/main/java/io/serverlessworkflow/impl/auth/DefaultAuthProviderFactory.java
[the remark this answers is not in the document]
fjtirado: I think all utility methods should be moved to a new static singleton class OauthUtils

fjtirado (commented):
@mcruzdev Lets move all utility methods to a new class OauthUtils. So basically remove the new class OauthPolicyData to OAuthUtils, and move the resolvePolicty method to there

impl/core/src/main/java/io/serverlessworkflow/impl/auth/OAuthPolicyData.java
```diff
@@ -0,0 +1,59 @@
...
+      OAuth2AuthenticationPolicyConfiguration config = oauth2.getOauth2();
+      return Optional.of(
+          new OAuthPolicyData(
+              config.getOAuth2ConnectAuthenticationProperties(),
+              config.getOAuth2AuthenticationPolicySecret(),
+              OAuthScheme.OAUTH2));
+    }
+    OpenIdConnectAuthenticationPolicy oidc = policy.getOpenIdConnectAuthenticationPolicy();
+    if (oidc != null && oidc.getOidc() != null) {
+      OpenIdConnectAuthenticationPolicyConfiguration config = oidc.getOidc();
```
fjtirado: I think that one of the benefits of using multipe return for this method  is that it allow to write thing like this
```suggestion
    if (oidc != null) {
      OpenIdConnectAuthenticationPolicyConfiguration config = oidc.getOidc();
      if (config != null) {
    }
```

impl/core/src/main/java/io/serverlessworkflow/impl/auth/OAuthPolicyData.java
```diff
@@ -0,0 +1,59 @@
...
+  }
+
+  public static Optional<OAuthPolicyData> from(AuthenticationPolicyUnion policy) {
+    if (policy == null) {
+      return Optional.empty();
+    }
+    OAuth2AuthenticationPolicy oauth2 = policy.getOAuth2AuthenticationPolicy();
+    if (oauth2 != null && oauth2.getOauth2() != null) {
+      OAuth2AuthenticationPolicyConfiguration config = oauth2.getOauth2();
+      return Optional.of(
```
fjtirado:
```suggestion
    if (oauth2 != null) {
      OAuth2AuthenticationPolicyConfiguration config = oauth2.getOauth2();
      if (config != null)  {
          return Optional.of(
```

impl/core/src/main/java/io/serverlessworkflow/impl/auth/OAuthUtils.java
fjtirado, line 1: This class should have an empty constructor and renamed to OAuthUtils

fjtirado (commented):
Also a very good one, just usual picky comments ;)

impl/core/src/main/java/io/serverlessworkflow/impl/auth/OAuthUtils.java
```diff
@@ -21,39 +21,68 @@
...
+
+  private OAuthUtils() {}
 
   public enum OAuthScheme {
     OAUTH2,
     OPENID_CONNECT
   }
 
+  public record OAuthPolicyData(
+      OAuth2AuthenticationData data, SecretBasedAuthenticationPolicy secret, OAuthScheme scheme) {}
```
fjtirado: hmmm, I think is better to not declare this as inner record, but in a separate class

impl/core/src/main/java/io/serverlessworkflow/impl/auth/OAuthUtils.java
```diff
@@ -21,39 +21,68 @@
...
+import io.serverlessworkflow.api.types.Workflow;
 import java.util.Optional;
 
-public record OAuthPolicyData(
-    OAuth2AuthenticationData data, SecretBasedAuthenticationPolicy secret, OAuthScheme scheme) {
+public class OAuthUtils {
+
+  private OAuthUtils() {}
 
   public enum OAuthScheme {
```
fjtirado: Same for this enum, it is bettes to have it in a separate class

impl/core/src/main/java/io/serverlessworkflow/impl/auth/OAuthUtils.java
```diff
@@ -21,39 +21,68 @@
...
+            new OAuthPolicyData(
+                config.getOpenIdConnectAuthenticationProperties(),
+                config.getOpenIdConnectAuthenticationPolicySecret(),
+                OAuthScheme.OPENID_CONNECT));
+      }
     }
     return Optional.empty();
   }
+
+  public static AuthenticationPolicyUnion resolvePolicy(
```
fjtirado: I do not know where this method is used, but should not return Optional.empty rather than null?

fjtirado (approved):

PR #1524, opened 2026-07-08
Replace serialization-based type inference with reified varargs type token

mdproctor:
## Summary

- Replace `SerializableFunction`/`SerializablePredicate`/`SerializableConsumer` + `ReflectionUtils.inferInputType()` with the **reified varargs type token** pattern across ~40 call sites
- Delete `ReflectionUtils`, `SerializableFunction`, `SerializablePredicate`, `SerializableConsumer` — no longer needed
- Drop `Serializable` from `InstanceIdFunction` and `UniqueIdBiFunction` (was only needed for the inference hack)
- Add test demonstrating nested generic type safety (`Map<String, List<Integer>>`)

## Motivation

The previous approach used JVM serialization internals (`writeReplace` → `SerializedLambda` → `getInstantiatedMethodType`) to recover erased type parameters at runtime. This is:

- **Fragile** — depends on undocumented JVM serialization behavior
- **Restrictive** — forces all lambdas to implement `Serializable`
- **Heavyweight** — serialization roundtrip + reflection per inference call

The reified varargs type token pattern achieves the same result (recovering the raw `Class<T>` at runtime) with zero reflection, zero serialization, and works with any lambda or method reference regardless of serializability. Java reifies array creation at the call site even for generic varargs parameters, so `T... typeToken` with zero arguments creates a `new T[0]` whose component type is recoverable via `getClass().getComponentType()`.

## Test plan

- [x] All existing tests pass (0 regressions)
- [x] New test: `function_with_nested_generics_preserves_type_safety` — verifies `Map<String, List<Integer>>` flows through with compile-time type safety
- [x] IntelliJ build: 0 errors


fjtirado:
@mdproctor 

We need serializable functions, consumer and predicates for  YAML/JSON serialization, so the serialization classes cannot be removed. They must be kept for Json/yaml serialization regardless of the inference mechanism used. 

For type inference, the PR proposal is smarter and faster than the current inference method based on SerializedLambda. But there is a caveat, it does not resolve the output class, which is internally used during execution time to convert the output.  I'm not sure the solution taken in AbstractJavaCallExecutor to circumvent that is correct  ( I guess the CI will fail). And even if it does not fail, the consequence of not knowing in advance that the function is returning a completable future is that we were creating an additional async request unnecessarily, which is probably heavier (I elaborate more in the last paragraph) than accessing the serializedlamda to check if the function is returning a CompletableFuture. 

That's said,  I really liked the approach for input parameter taken here, and since we need SerializedLambda for YAML/Json serialization, I'm thinking on an hybrid approach. I was already contemplating some extra improvements on ReflectionUtils (calling inferMethodType only once and cache the resolution of Method Type) which, together with the component type approach might dramatically improve type inference performance without losing the completablefuture check at runtime. 

An important remark about inference performance, the type inference only occurs during workflow definition time (typically at startup), so the performance improvement does not really affect execution time, but startup time, since the definitions are loaded at startup, hence we do not invest much effort on improving type inference performance (in other words, type inference is a one time operation, that is the reason why the CompletablueFuture change is important, it affects all workflow executions, while output class inference is just performed once during workflow definition, so it can be argued that the current change in AbstractJavaCallExceutor will make performance worse when invoking java methods that are already async, as paradoxically as it might sound)

ricardozanini:
PR for reference only, thanks Mark!

PR #1551, opened 2026-07-20
Initialize OAuth2 default endpoints and expose WorkflowDefinitionId separator

ricardozanini:
## Summary

- Initialize `OAuth2AuthenticationPropertiesEndpoints` in `OIDCBuilder` constructor so OAuth2 objects created via the fluent API have spec-compliant default endpoints (`/oauth2/token`, `/oauth2/revoke`, `/oauth2/introspect`) instead of `null`
- Add `DEFAULT_SEPARATOR` constant (`:`) to `WorkflowDefinitionId` and use it in `toString()`, so downstream projects (e.g. quarkus-flow) can reference the canonical separator without duplicating it

## Test plan

- [x] All 162 tests in `fluent/spec` and `impl/core` pass
- [ ] Verify OAuth2 endpoints are non-null when building via fluent API
- [ ] Verify `WorkflowDefinitionId.DEFAULT_SEPARATOR` is accessible from downstream

### DIFF
--- a/fluent/spec/src/main/java/io/serverlessworkflow/fluent/spec/OAuth2AuthenticationPolicyBuilder.java
+++ b/fluent/spec/src/main/java/io/serverlessworkflow/fluent/spec/OAuth2AuthenticationPolicyBuilder.java
@@ -25,6 +25,7 @@ public final class OAuth2AuthenticationPolicyBuilder
 
   OAuth2AuthenticationPolicyBuilder() {
     super();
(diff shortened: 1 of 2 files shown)


fjtirado (approved):

fjtirado:
@ricardozanini I cannot merge till you sign the PR (yes, its something new needed after repo move that can be configured in the IDE in 5 minutes)

PR #1552, opened 2026-07-22
Remove CallableTaskFactory indirection from CallableTaskBuilder

mdproctor:
## Summary

- Replaces the two-step `init()` → `CallableTaskFactory.get()` pattern on `CallableTaskBuilder` with a single `build()` method that returns `CallableTask` directly
- Deletes the `CallableTaskFactory` interface from the public API
- Updates all six implementations and the `CallTaskExecutor` call site

## Motivation

`CallableTaskFactory` is a `Supplier<CallableTask>` whose `.get()` is called immediately after `init()` returns — no implementation uses the deferred construction. Every implementation already captures all state in the lambda closure returned from `init()`, so the factory layer adds a public type, an allocation, and a concept to the API surface with no behavioral benefit.

After this change:
- **Smaller public API** — one fewer type for implementors to learn and maintain. Third-party `CallableTaskBuilder` implementations (e.g. in [Quarkus Flow](https://github.com/quarkiverse/quarkus-flow)) drop the factory wrapper and return `CallableTask` directly.
- **Simpler call site** — `CallTaskExecutorBuilder` stores a `CallableTask` directly instead of holding a factory and deferring `.get()`.
- **No behavioral change** — same data in, same `CallableTask` out. All existing tests pass.

## Test plan

- [x] `mvn verify` — all 232 tests pass (one pre-existing flaky `SchedulerTest.testAfter` timing test passes on rerun)
- [x] No remaining references to `CallableTaskFactory` in the codebase


fjtirado:
The two step is needed for CallFunction (so all functions are really defined when the second step is initiated)

PR #1556, opened 2026-07-22
Add defaultEventSource to WorkflowApplication

### LINKED ISSUE #1554: Improve CloudEvent defaults and validation for emit events
## Problem

When emitting events via `emit.event.with`, the [spec requires](https://github.com/open-workflow-specification/specification/blob/main/dsl-reference.md#event-properties) `id`, `source`, and `type` to be set. The current implementation has gaps:

1. **No model-level validation** — `EventProperties` has zero `@NotNull`/`@NotBlank` annotations. All fields are optional at the type level, with enforcement only at runtime in `EmitExecutor`.
2. **Weak `source` default** — `CloudEventUtils.source()` returns `URI.create("reference-impl")`, which provides no traceability.
3. **No `time` default** — emitted events have no timestamp unless the user explicitly sets one. CloudEvents best practice recommends always including `time`.

## Current runtime behavior (`EmitExecutor`)

| Field | Behavior when missing |
|-------|----------------------|
| `id` | Auto-generates UUID — good |
| `source` | Falls back to `"reference-impl"` — weak |
| `type` | Throws `IllegalArgumentException` — correct |
| `time` | Not set — should default to now |

### 1. Improve `CloudEventUtils.source()` default
Replace `"reference-impl"` with something meaningful. Options:
- `"io.serverlessworkflow.sdk-java"` (static, identifies the SDK)
- Derive from `WorkflowDefinitionId`: `"{namespace}/{name}:{version}"` (e.g., `"org-acme/order-processor:1.0.0"`)

The second option requires passing context into `EmitExecutor.buildCloudEvent()`, which already has access to `WorkflowContext`.

### 2. Default `time` to `OffsetDateTime.now()` in `EmitExecutor`
When `timeFilter` is empty, set `ceBuilder.withTime(OffsetDateTime.now())` instead of omitting it.

### 3. Add defaults in the DSL layer (`EmitSpec` / `produced()`)
The `EmitSpec` convenience methods could embed defaults so the common path produces valid, traceable CloudEvents:
- `produced(type)` could auto-set `source` (e.g., from a configurable default)
- `randomId()` evaluated at definition time won't work for `id` (needs to be unique per execution), so the runtime fallback is the right place

### 4. Make `source` configurable at `WorkflowApplication` level
Allow setting a default source once that applies to all emitted events, overridable per-emit.

### Note on validation annotations
`EventProperties` is shared between emit and filter contexts (where requirements differ), so adding `@NotNull` directly would break filter use cases. Runtime validation in `EmitExecutor` is the correct approach unless a separate emit-specific type is introduced.

matheusandre1:

**What this PR does / why we need it**: Closes: #1554 

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
@@ -30,6 +30,7 @@ import io.serverlessworkflow.impl.config.SecretManager;
 import io.serverlessworkflow.impl.config.SystemPropertyConfigManager;
 import io.serverlessworkflow.impl.events.CloudEventPredicateFactory;
 import io.serverlessworkflow.impl.events.DefaultCloudEventPredicateFactory;
+import io.serverlessworkflow.impl.events.EmitSourceResolver;
 import io.serverlessworkflow.impl.events.EventConsumer;
 import io.serverlessworkflow.impl.events.EventPublisher;
 import io.serverlessworkflow.impl.events.InMemoryEvents;
@@ -105,6 +106,7 @@ public class WorkflowApplication implements AutoCloseable {
   private final Optional<URITemplateResolver> templateResolver;
   private final Optional<FunctionReader> functionReader;
   private final URI defaultCatalogURI;
+  private final WorkflowValueResolver<URI> defaultEventSource;
   private final Collection<CallableTaskProxyBuilder> callableProxyBuilders;
   private final CloudEventPredicateFactory cloudEventPredicateFactory;
   private final AllStrategyCorrelationInfoFactory allStrategyCorrelationInfoFactory;
@@ -138,6 +140,7 @@ public class WorkflowApplication implements AutoCloseable {
     this.templateResolver = builder.templateResolver;
     this.functionReader = builder.functionReader;
     this.defaultCatalogURI = builder.defaultCatalogURI;
+    this.defaultEventSource = builder.defaultEventSource;
     this.id = builder.id;
     this.callableProxyBuilders = builder.callableProxyBuilders;
     this.cloudEventPredicateFactory = builder.cloudEventPredicateFactory;
@@ -263,6 +266,7 @@ public class WorkflowApplication implements AutoCloseable {
     private Optional<URITemplateResolver> templateResolver;
     private Optional<FunctionReader> functionReader;
     private URI defaultCatalogURI;
+    private WorkflowValueResolver<URI> defaultEventSource;
     private CloudEventPredicateFactory cloudEventPredicateFactory;
     private AllStrategyCorrelationInfoFactory allStrategyCorrelationInfoFactory;
     private WorkflowLifeCycleCloudEventFactory lifeCycleCloudEventFactory;
@@ -404,6 +408,15 @@ public class WorkflowApplication implements AutoCloseable {
       return this;
     }
 
+    public Builder withDefaultEventSource(URI defaultEventSource) {
+      return withDefaultEventSource((workflow, task, model) -> defaultEventSource);
+    }
+
+    public Builder withDefaultEventSource(WorkflowValueResolver<URI> defaultEventSource) {
+      this.defaultEventSource = defaultEventSource;
+      return this;
+    }
+
     public Builder withCloudEventPredicateFactory(
         CloudEventPredicateFactory cloudEventPredicateFactory) {
       this.cloudEventPredicateFactory = cloudEventPredicateFactory;
@@ -521,6 +534,9 @@ public class WorkflowApplication implements AutoCloseable {
       if (defaultCatalogURI == null) {
         defaultCatalogURI = URI.create("https://github.com/serverlessworkflow/catalog");
       }
+      if (defaultEventSource == null) {
+        defaultEventSource = new EmitSourceResolver();
+      }
       Collections.sort(listeners);
       Collections.sort(callableProxyBuilders);
       if (id == null) {
@@ -637,6 +653,10 @@ public class WorkflowApplication implements AutoCloseable {
 
   public URI defaultCatalogURI() {
     return defaultCatalogURI;
+  }
+
+  public WorkflowValueResolver<URI> defaultEventSource() {
+    return defaultEventSource;
   }
 
   public String id() {
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/events/EmitSourceResolver.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/events/EmitSourceResolver.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2020-Present The Serverless Workflow Specification Authors
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package io.serverlessworkflow.impl.events;
+
+import io.serverlessworkflow.impl.TaskContext;
+import io.serverlessworkflow.impl.WorkflowContext;
+import io.serverlessworkflow.impl.WorkflowModel;
+import io.serverlessworkflow.impl.WorkflowValueResolver;
+import java.net.URI;
+
+public class EmitSourceResolver implements WorkflowValueResolver<URI> {
+
+  private static final String SEP = "/";
+
+  @Override
+  public URI apply(WorkflowContext workflow, TaskContext task, WorkflowModel model) {
+    return URI.create(
+        SEP
+            + workflow.definition().application().id()
+            + SEP
+            + workflow.definition().id().toString(SEP));
+  }
+}
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/executors/EmitExecutor.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/executors/EmitExecutor.java
@@ -100,17 +100,24 @@ public class EmitExecutor extends RegularTaskExecutor<EmitTask> {
             .sourceFilter()
             .map(filter -> filter.apply(workflow, taskContext, taskContext.input()))
             .map(URI::create)
-            .orElse(CloudEventUtils.source()));
+            .orElseGet(
+                () ->
+                    workflow
+                        .definition()
+                        .application()
+                        .defaultEventSource()
+                        .apply(workflow, taskContext, taskContext.input())));
     ceBuilder.withType(
         props
             .typeFilter()
             .map(filter -> filter.apply(workflow, taskContext, taskContext.input()))
             .orElseThrow(
                 () -> new IllegalArgumentException("Type is required for emitting events")));
-    props
-        .timeFilter()
-        .map(filter -> filter.apply(workflow, taskContext, taskContext.input()))
-        .ifPresent(value -> ceBuilder.withTime(value));
+    ceBuilder.withTime(
+        props
+            .timeFilter()
+            .map(filter -> filter.apply(workflow, taskContext, taskContext.input()))
+            .orElseGet(OffsetDateTime::now));
     props
         .subjectFilter()
         .map(filter -> filter.apply(workflow, taskContext, taskContext.input()))
--- a/impl/test/src/test/java/io/serverlessworkflow/impl/test/EmitExecutorTest.java
+++ b/impl/test/src/test/java/io/serverlessworkflow/impl/test/EmitExecutorTest.java
@@ -1,0 +1,200 @@
+/*
+ * Copyright 2020-Present The Serverless Workflow Specification Authors
+ *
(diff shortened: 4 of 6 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/executors/EmitExecutor.java
```diff
@@ -100,17 +100,19 @@ private CloudEvent buildCloudEvent(WorkflowContext workflow, TaskContext taskCon
...
                 () -> new IllegalArgumentException("Type is required for emitting events")));
-    props
-        .timeFilter()
-        .map(filter -> filter.apply(workflow, taskContext, taskContext.input()))
-        .ifPresent(value -> ceBuilder.withTime(value));
+    ceBuilder.withTime(
+        props
+            .timeFilter()
+            .map(filter -> filter.apply(workflow, taskContext, taskContext.input()))
+            .orElseGet(OffsetDateTime::now));
```
ricardozanini, line 120: Do we have test coverage for this?

ricardozanini:
@matheusandre1 you must sign your commits with `-S`.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/EmitExecutor.java
matheusandre1: > Do we have test coverage for this?

no, Should I implement it? Right?

matheusandre1:
> @matheusandre1 you must sign your commits with `-S`.

Sorry, I always forget about that.

fjtirado (commented):
@matheusandre1 
Lets make default event source a WorkflowValueResolver<URI>, so it returns an URI from the WorkflowContext, TaskContext and Model

impl/core/src/main/java/io/serverlessworkflow/impl/events/EmitSourceResolver.java
```diff
@@ -0,0 +1,35 @@
...
+import io.serverlessworkflow.impl.WorkflowContext;
+import io.serverlessworkflow.impl.WorkflowDefinitionId;
+import io.serverlessworkflow.impl.WorkflowModel;
+import io.serverlessworkflow.impl.WorkflowValueResolver;
+import java.net.URI;
+
+public class EmitSourceResolver implements WorkflowValueResolver<URI> {
+
+  private static final String SEP = "/";
+  private static final String VER_SEP = ":";
```
fjtirado: @ricardozanini why a ":" and not a "/"

I think a calculated default source should combine applicationid and definition id and not use any other separator than /, so, in my opinion the code should be


`workflow.definition().application().id()+SEP+workflow.definition().id().toString(SEP);`

You can remove VER_SEP

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -263,6 +266,7 @@ public SchemaValidator getValidator(SchemaInline inline) {
     private Optional<URITemplateResolver> templateResolver;
     private Optional<FunctionReader> functionReader;
     private URI defaultCatalogURI;
+    private WorkflowValueResolver<URI> defaultEventSource = new EmitSourceResolver();
```
fjtirado: This should be initialized only if withDefaultEventSource has not being invoked.

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -404,6 +408,19 @@ public Builder withDefaultCatalogURI(URI defaultCatalogURI) {
       return this;
     }
 
+    public Builder withDefaultEventSource(String defaultEventSource) {
+      return withDefaultEventSource(URI.create(defaultEventSource));
```
fjtirado: If the string is not a valir URI, application initialization will fail. Im not sure we should have this method to be hones

impl/test/src/test/java/io/serverlessworkflow/impl/test/EmitExecutorTest.java
```diff
@@ -0,0 +1,198 @@
...
+    List<CloudEvent> events = Collections.synchronizedList(new ArrayList<>());
+    InMemoryEvents broker = new InMemoryEvents();
+    broker.register(eventType, events::add);
+    WorkflowApplication app =
+        WorkflowApplication.builder().withEventConsumer(broker).withEventPublisher(broker).build();
+    try {
+      app.workflowDefinition(wf).instance().start().join();
+    } finally {
+      app.close();
+    }
```
fjtirado:
```suggestion
    try (WorkflowApplication app =
        WorkflowApplication.builder().withEventConsumer(broker).withEventPublisher(broker).build())
    {
      app.workflowDefinition(wf).instance().start().join();
    }
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/EmitExecutorTest.java
```diff
@@ -0,0 +1,198 @@
...
+    InMemoryEvents broker = new InMemoryEvents();
+    broker.register("org.example.event", events::add);
+    WorkflowApplication app =
+        WorkflowApplication.builder()
+            .withDefaultEventSource("app.source")
+            .withEventConsumer(broker)
+            .withEventPublisher(broker)
+            .build();
+    app.workflowDefinition(wf).instance().start().join();
+    app.close();
```
fjtirado:
```suggestion
    try (WorkflowApplication app =
        WorkflowApplication.builder()
            .withDefaultEventSource("app.source")
            .withEventConsumer(broker)
            .withEventPublisher(broker)
            .build()) {
    app.workflowDefinition(wf).instance().start().join();
    }
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/EmitExecutorTest.java
```diff
@@ -0,0 +1,198 @@
...
+    List<CloudEvent> events = Collections.synchronizedList(new ArrayList<>());
+    InMemoryEvents broker = new InMemoryEvents();
+    broker.register("org.example.event", events::add);
+    WorkflowApplication app =
+        WorkflowApplication.builder()
+            .withDefaultEventSource(
+                (workflow, task, model) -> URI.create("apps/" + workflow.definition().id().name()))
+            .withEventConsumer(broker)
+            .withEventPublisher(broker)
+            .build();
```
fjtirado: Same here, apparently AI does not know about try {} existance

impl/test/src/test/java/io/serverlessworkflow/impl/test/EmitExecutorTest.java
```diff
@@ -0,0 +1,198 @@
...
+import java.util.List;
+import java.util.function.Consumer;
+import org.junit.jupiter.api.Test;
+
+class EmitExecutorTest {
+
+  private static List<CloudEvent> runWorkflow(Workflow wf, String eventType) {
+    List<CloudEvent> events = Collections.synchronizedList(new ArrayList<>());
+    InMemoryEvents broker = new InMemoryEvents();
+    broker.register(eventType, events::add);
```
fjtirado: There is a lot of duplicated code in these test, lets avoid copy paste and use "@beforeeach" annotation

impl/core/src/main/java/io/serverlessworkflow/impl/events/EmitSourceResolver.java
ricardozanini: Yes, I agree. Not sure where the `:` came from. Yes, it must be a valid URI; see: https://github.com/cloudevents/spec/blob/v1.0.2/cloudevents/spec.md#source-1

@matheusandre1 can you change it, please?

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
ricardozanini: Yeah, it must fail fast.

matheusandre1: > If the string is not a valir URI, application initialization will fail. Im not sure we should have this method to be hones

I'm a little confused here, should I remove it or keep it?

fjtirado: lets keep it

ricardozanini: @fjtirado we must have somethign for the downstream client to implement/apply as a supplier. For instance, in Quarkus Flow we want to set the default source to `/{namespace}/{name}/{version}`.

fjtirado: yes, thats the WorkflowValueResolver<URI) that accepts the three parameters. This is a particular case where the source is a constant string which Im not sure is useful and is erro prompt if the string cannot be converted to an URI (note that I did not question the constant URI one)

ricardozanini: Ok makes sense removing it then.

ricardozanini:
@matheusandre1 when generating the `source` URI please add the prefix `/`. E.g. `/org-acme/payment-processor/1.0.0`

impl/core/src/main/java/io/serverlessworkflow/impl/events/EmitSourceResolver.java
```diff
@@ -0,0 +1,42 @@
...
+        SEP
+            + workflow.definition().application().id()
+            + SEP
+            + workflow.definition().id().toString(SEP);
+    try {
+      // Use the multi-arg URI constructor to ensure proper escaping of unsafe characters.
+      return new URI(null, null, path, null, null);
+    } catch (java.net.URISyntaxException e) {
+      throw new IllegalArgumentException("Invalid default event source: " + path, e);
+    }
```
fjtirado: why do we need to escape unsafe characters here?
If this change is result of an AI comment and not of a test execution, lets switch back to URI.create(), which already thrown IllegalArgument exception so you do not need to try---catch either.

impl/core/src/main/java/io/serverlessworkflow/impl/events/EmitSourceResolver.java
```diff
@@ -0,0 +1,33 @@
...
+
+public class EmitSourceResolver implements WorkflowValueResolver<URI> {
+
+  private static final String SEP = "/";
+
+  @Override
+  public URI apply(WorkflowContext workflow, TaskContext task, WorkflowModel model) {
+    return URI.create(
+        workflow.definition().application().id() + SEP + workflow.definition().id().toString(SEP));
+  }
```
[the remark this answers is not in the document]
fjtirado: NO, this coments is plainly stupid that makes the code unnecesarily ugly, you should have kept the original. Source is an URI for convention, this is not going to cause the Cloud Event (tipically a json) to not be published

impl/test/src/test/java/io/serverlessworkflow/impl/test/EmitExecutorTest.java
```diff
@@ -0,0 +1,199 @@
...
+    Workflow wf = emitWorkflow("emit-time-default", "org.example.event");
+    List<CloudEvent> events = runWorkflow(wf, "org.example.event");
+
+    assertThat(events).hasSize(1);
+    CloudEvent ev = events.get(0);
+    OffsetDateTime now = OffsetDateTime.now();
+    assertThat(ev.getTime()).isNotNull();
+    assertThat(ev.getTime()).isBeforeOrEqualTo(now);
+    assertThat(ev.getTime()).isAfter(now.minusSeconds(30));
+  }
```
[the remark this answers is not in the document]
fjtirado: This comment however makes sense (but the code suggested can be enhanced), you should capture the time before the workflow is executed and ensure now is just after the time you capture

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -404,6 +408,15 @@ public Builder withDefaultCatalogURI(URI defaultCatalogURI) {
...
     }
 
+    public Builder withDefaultEventSource(URI defaultEventSource) {
+      return withDefaultEventSource((workflow, task, model) -> defaultEventSource);
+    }
+
+    public Builder withDefaultEventSource(WorkflowValueResolver<URI> defaultEventSource) {
+      this.defaultEventSource = defaultEventSource;
+      return this;
+    }
```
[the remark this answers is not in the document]
fjtirado: Just ignor that, if the parameter is null, the default will be used
The IA should have realize that.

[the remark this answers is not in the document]
matheusandre1: > Just ignor that, if the parameter is null, the default will be used The IA should have realize that.

Hahaha, I didn't do that with that—unfortunately, it was the human here.

ricardozanini (approved):

fjtirado (approved):

ricardozanini:
@matheusandre1 wanna open the counterpart in quarkus-flow? https://github.com/quarkiverse/quarkus-flow/issues/776 We can use the defaultSource from here.

matheusandre1:
> @matheusandre1 wanna open the counterpart in quarkus-flow? [quarkiverse/quarkus-flow#776](https://github.com/quarkiverse/quarkus-flow/issues/776) We can use the defaultSource from here.

I would, actually, but there are two open pull requests mentioning the issue.

matheusandre1:
The GitHub PR is down; I'll submit the pull request when the next SDK release comes out.
cc: @ricardozanini

PR #1568, opened 2026-07-23
Improve CloudEvent defaults for emitted events (#1554)

### LINKED ISSUE #1554: Improve CloudEvent defaults and validation for emit events

anuragpaul602-netizen:
## What & why

Fixes #1554.

Emitted CloudEvents (`emit.event.with`) were under-specified relative to the spec, which requires `id`, `source`, and `type`:

| Field | Before | After |
|-------|--------|-------|
| `id` | auto UUID — good | unchanged |
| `type` | throws if missing — correct | unchanged |
| `source` | falls back to `reference-impl` — no traceability | derived from workflow identity / configurable |
| `time` | omitted entirely | defaults to `OffsetDateTime.now()` |

## How

- **`source`** — when an emit declares no source, derive it from the emitting workflow's identity, `namespace:name:version`, via a new `CloudEventUtils.source(WorkflowDefinitionId)`. `EmitExecutor.buildCloudEvent` already has the `WorkflowContext`, so no new plumbing is needed. The no-arg static fallback changes from `reference-impl` to `urn:serverlessworkflow:sdk-java`.
- **`time`** — in `EmitExecutor`, when the `timeFilter` is empty, set `withTime(OffsetDateTime.now())` instead of omitting the attribute. Explicit literal/expression `time` is untouched.
- **Application-level default source** — new `WorkflowApplication.Builder#withDefaultEventSource(URI|String)` lets an application set one source for all emitted events.
- **Resolution order** (highest wins): explicit per-emit `source` → application-level default → workflow-identity-derived → static fallback.

### Validation

`EventProperties` is deliberately left without `@NotNull`/`@NotBlank` annotations: it is shared between the *emit* and *filter* contexts, where required-ness differs, so annotating it would break filter use. With `id`, `source`, and `time` now all defaulted, `type` remains the only required emit attribute and `EmitExecutor` already throws when it is missing — runtime enforcement remains the correct approach.

## Compatibility note

Changing the default `source` from `reference-impl` is a behavioral change; downstream consumers that filtered or asserted on `source == "reference-impl"` will need to update. This seems acceptable pre-1.0 but is called out here for release notes.

## Testing

`EmitCloudEventDefaultsTest` (new) registers a capturing `EventPublisher` and asserts:
- an emit with no source gets `source == namespace:name:version` and a `time` close to now;
- an application-level default source overrides the identity-derived one;
- an explicit per-emit source overrides the application default.

Full `impl/test` + `impl/lifecycleevent` suites pass (239 tests), confirming no regression in lifecycle-event or correlation behavior.

## Downstream

Quarkus Flow tracks the DSL-ergonomics side in quarkiverse/quarkus-flow#776.

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
@@ -110,6 +110,7 @@ public class WorkflowApplication implements AutoCloseable {
   private final AllStrategyCorrelationInfoFactory allStrategyCorrelationInfoFactory;
   private final WorkflowLifeCycleCloudEventFactory lifeCycleCloudEventFactory;
   private final ScheduledExecutorService schedulerExecutorService;
+  private final Optional<URI> defaultEventSource;
 
   private WorkflowApplication(Builder builder) {
     this.taskFactory = builder.taskFactory;
@@ -144,6 +145,18 @@ public class WorkflowApplication implements AutoCloseable {
     this.allStrategyCorrelationInfoFactory = builder.allStrategyCorrelationInfoFactory;
     this.lifeCycleCloudEventFactory = builder.lifeCycleCloudEventFactory;
     this.schedulerExecutorService = builder.schedulerExecutorService;
+    this.defaultEventSource = builder.defaultEventSource;
+  }
+
+  /**
+   * Optional application-wide default CloudEvent {@code source} for emitted events. When set, it
+   * takes precedence over the workflow-identity-derived source but is still overridden by an
+   * explicit per-emit {@code source}.
+   *
+   * @return the configured default source, or empty if none was set
+   */
+  public Optional<URI> defaultEventSource() {
+    return defaultEventSource;
   }
 
   public TaskExecutorFactory taskFactory() {
@@ -248,6 +261,7 @@ public class WorkflowApplication implements AutoCloseable {
     private ExecutorServiceFactory executorFactory = new DefaultExecutorServiceFactory();
     private EventConsumer<?, ?> eventConsumer;
     private Collection<EventPublisher> eventPublishers = new ArrayList<>();
+    private Optional<URI> defaultEventSource = Optional.empty();
     private RuntimeDescriptorFactory descriptorFactory =
         () -> new RuntimeDescriptor("reference impl", "1.0.0_alpha", Collections.emptyMap());
     private boolean lifeCycleCEPublishingEnabled = true;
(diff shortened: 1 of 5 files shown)


fjtirado:
This is a duplicate of https://github.com/open-workflow-specification/sdk-java/pull/1556

PR #1594, opened 2026-08-03
Guard sdk-java from NoSuchElementException

### LINKED ISSUE #1593: WorkflowDefinition crashes when workflow has no tasks
Creating a WorkflowDefinition from a Workflow with an empty task list causes a crash.

Steps to reproduce:
```java
Workflow workflow = WorkflowBuilder.workflow().tasks().build();
WorkflowDefinition.of(application, workflow);
```
Actual behavior:

TaskExecutorHelper.createExecutorList() calls iterator.next() without checking hasNext() on the empty task map, resulting in a NoSuchElementException.

mcruzdev:

Closes #1593

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowMutableInstance.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowMutableInstance.java
@@ -18,7 +18,6 @@ package io.serverlessworkflow.impl;
 import static io.serverlessworkflow.impl.LifecycleEventsUtils.publishEvent;
 import static io.serverlessworkflow.impl.WorkflowUtils.validationError;
 
-import io.serverlessworkflow.impl.executors.TaskExecutor;
 import io.serverlessworkflow.impl.executors.TaskExecutorHelper;
 import io.serverlessworkflow.impl.lifecycle.WorkflowCancelledEvent;
 import io.serverlessworkflow.impl.lifecycle.WorkflowCompletedEvent;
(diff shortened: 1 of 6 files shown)


fjtirado (changes requested):
Hmmmm, I need a yaml reproducer for this one. If there is not such reproducer, then the fix should be in the fluent dsl cpde, to prevent a not valid workflow definition to be generated rather that adding defensive checks in the implementation code

If, as I suspected, there is not way that the YAML file is validated with an empty task list, then the solution should be to do a check in the build method of the fluent DSL to prevent that. Or, in other word, there should be not workflow object created with empty task list. And it is is, it is perfectly ok for the implementation code to throw any unexpected exception, because the workflow object is not a valid one.

mcruzdev:
> Or, in other word, there should be not workflow object created with empty task list. And it is is, it is perfectly ok for the implementation code to throw any unexpected exception, because the workflow object is not a valid one.

I prefer to throw an exception centralizing the validation and additionally to create an empty list when there is no task added while using the DSL.

fjtirado:
> > Or, in other word, there should be not workflow object created with empty task list. And it is is, it is perfectly ok for the implementation code to throw any unexpected exception, because the workflow object is not a valid one.
> 
> I prefer to throw an exception centralizing the validation

Im not fan of double validation (which is what you are doing in this PR, because the yaml is already validated) this "validation" should be done in the DSL.

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowMutableInstance.java
```diff
@@ -117,6 +109,17 @@ protected final CompletableFuture<WorkflowModel> startExecution(
...
+    WorkflowModel inputModel =
+        workflowContext
+            .definition()
+            .inputFilter()
+            .map(f -> f.apply(workflowContext, null, input))
+            .orElse(input);
+    return TaskExecutorHelper.processTaskList(
+        workflowContext.definition().startTask(), workflowContext, Optional.empty(), inputModel);
+  }
+
```
fjtirado: Is this really changing something or it is just a not needed refactor?

mcruzdev: I will remove this one in the next commit, I will add a validation at build method.

fjtirado (approved):

PR #1595, opened 2026-08-04
Move jackson native file to jackson module

fjtirado:
Although the current approach work, it is better to have each native file in each module


gmunozfe (approved):
LGTM, well done

PR #1600, opened 2026-08-05
Implement retry attempt.duration per-attempt timeout

### LINKED ISSUE #1526: Implement attemp.duration
The current text spec is kind of confusing regarding max duration attemtp for retries.  
https://github.com/serverlessworkflow/specification/blob/main/dsl-reference.md#retry-limit

Once it is clarified, we should proceeed implementing it.

mcruzdev:
Add support for `attempt.duration` per attempt timeout.

Closes #1526 

### DIFF
--- a/impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
+++ b/impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
@@ -59,6 +59,7 @@ public class TryExecutor extends RegularTaskExecutor<TryTask> {
   private final Optional<TaskExecutor<?>> catchTaskExecutor;
   private final Optional<RetryExecutor> retryIntervalExecutor;
   private final Optional<WorkflowValueResolver<Duration>> attemptDuration;
+  private final Optional<WorkflowValueResolver<Duration>> overallDuration;
   private final String errorVariable;
 
   public static class TryExecutorBuilder extends RegularTaskExecutorBuilder<TryTask> {
@@ -70,6 +71,7 @@ public class TryExecutor extends RegularTaskExecutor<TryTask> {
     private final Optional<TaskExecutor<?>> catchTaskExecutor;
     private final Optional<RetryExecutor> retryIntervalExecutor;
     private final Optional<WorkflowValueResolver<Duration>> attemptDuration;
+    private final Optional<WorkflowValueResolver<Duration>> overallDuration;
     private String errorVariable;
 
     protected TryExecutorBuilder(
@@ -89,31 +91,33 @@ public class TryExecutor extends RegularTaskExecutor<TryTask> {
                       position.copy().addProperty("catch"), catchTaskDo, definition))
               : Optional.empty();
       Retry retry = catchInfo.getRetry();
-      Optional<RetryPolicy> retryPolicy =
-          retry != null ? resolveRetryPolicy(retry) : Optional.empty();
+      Optional<RetryPolicy> retryPolicy = resolveRetryPolicy(retry);
       this.retryIntervalExecutor = retryPolicy.map(this::buildRetryExecutor);
       this.attemptDuration = retryPolicy.flatMap(this::resolveAttemptDuration);
+      this.overallDuration = retryPolicy.flatMap(this::resolveOverallDuration);
       this.taskExecutor =
           TaskExecutorHelper.createExecutorList(position, task.getTry(), definition, "try");
     }
 
     private Optional<RetryPolicy> resolveRetryPolicy(Retry retry) {
-      if (retry.getRetryPolicyDefinition() != null) {
-        return Optional.of(retry.getRetryPolicyDefinition());
-      } else if (retry.getRetryPolicyReference() != null) {
-        RetryPolicy retryPolicy =
-            workflow
-                .getUse()
-                .getRetries()
-                .getAdditionalProperties()
-                .get(retry.getRetryPolicyReference());
-        if (retryPolicy == null) {
-          throw new IllegalStateException(
-              "Retry policy " + retry.getRetryPolicyReference() + " was not found");
+      RetryPolicy retryPolicy = null;
+      if (retry != null) {
+        if (retry.getRetryPolicyDefinition() != null) {
+          retryPolicy = retry.getRetryPolicyDefinition();
+        } else if (retry.getRetryPolicyReference() != null) {
+          retryPolicy =
+              workflow
+                  .getUse()
+                  .getRetries()
+                  .getAdditionalProperties()
+                  .get(retry.getRetryPolicyReference());
+          if (retryPolicy == null) {
+            throw new IllegalStateException(
+                "Retry policy " + retry.getRetryPolicyReference() + " was not found");
+          }
         }
-        return Optional.of(retryPolicy);
       }
-      return Optional.empty();
+      return Optional.ofNullable(retryPolicy);
     }
 
     protected RetryExecutor buildRetryExecutor(RetryPolicy retryPolicy) {
@@ -130,6 +134,14 @@ public class TryExecutor extends RegularTaskExecutor<TryTask> {
       return limit != null && limit.getAttempt() != null && limit.getAttempt().getDuration() != null
           ? Optional.of(
               WorkflowUtils.fromTimeoutAfter(application, limit.getAttempt().getDuration()))
+          : Optional.empty();
+    }
+
+    private Optional<WorkflowValueResolver<Duration>> resolveOverallDuration(
+        RetryPolicy retryPolicy) {
+      RetryLimit limit = retryPolicy.getLimit();
+      return limit != null && limit.getDuration() != null
+          ? Optional.of(WorkflowUtils.fromTimeoutAfter(application, limit.getDuration()))
           : Optional.empty();
     }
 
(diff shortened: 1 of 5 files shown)


impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
```diff
@@ -83,27 +89,32 @@ protected TryExecutorBuilder(
...
         if (retryPolicy == null) {
-          throw new IllegalStateException("Retry policy " + retryPolicy + " was not found");
+          throw new IllegalStateException(
+              "Retry policy " + retry.getRetryPolicyReference() + " was not found");
         }
+        return retryPolicy;
       }
-      return retryPolicy != null ? Optional.of(buildRetryExecutor(retryPolicy)) : Optional.empty();
+      return null;
     }
```
fjtirado: I do not get that change

fjtirado: Ok, now I get it, but it not good, in my opinion. 
I think this should still return an optional and then use the optional to build the two other optional (map method is your friend) rather than starting a null check party in the caller.

fjtirado (commented):
Refactor using optional rather than null

impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
```diff
@@ -114,6 +125,16 @@ protected RetryExecutor buildRetryExecutor(RetryPolicy retryPolicy) {
...
 
+    private Optional<WorkflowValueResolver<Duration>> resolveAttemptDuration(
+        RetryPolicy retryPolicy) {
+      RetryLimit limit = retryPolicy.getLimit();
+      if (limit != null && limit.getAttempt() != null && limit.getAttempt().getDuration() != null) {
+        return Optional.of(
+            WorkflowUtils.fromTimeoutAfter(application, limit.getAttempt().getDuration()));
+      }
+      return Optional.empty();
```
fjtirado:
```suggestion
      return limit != null && limit.getAttempt() != null && limit.getAttempt().getDuration() != null ?
        return Optional.of(
            WorkflowUtils.fromTimeoutAfter(application, limit.getAttempt().getDuration())) :
      return Optional.empty();
```

impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
```diff
@@ -83,27 +89,32 @@ protected TryExecutorBuilder(
                       position.copy().addProperty("catch"), catchTaskDo, definition))
               : Optional.empty();
       Retry retry = catchInfo.getRetry();
-      this.retryIntervalExecutor = retry != null ? buildRetryInterval(retry) : Optional.empty();
+      RetryPolicy retryPolicy = retry != null ? resolveRetryPolicy(retry) : null;
+      this.retryIntervalExecutor =
+          retryPolicy != null ? Optional.of(buildRetryExecutor(retryPolicy)) : Optional.empty();
+      this.attemptDuration =
+          retryPolicy != null ? resolveAttemptDuration(retryPolicy) : Optional.empty();
```
fjtirado: This you should refactor  using Optional.map when resolveRetryPolicy return optional

impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
```diff
@@ -83,27 +89,32 @@ protected TryExecutorBuilder(
...
+        RetryPolicy retryPolicy =
             workflow
                 .getUse()
                 .getRetries()
                 .getAdditionalProperties()
                 .get(retry.getRetryPolicyReference());
         if (retryPolicy == null) {
-          throw new IllegalStateException("Retry policy " + retryPolicy + " was not found");
+          throw new IllegalStateException(
+              "Retry policy " + retry.getRetryPolicyReference() + " was not found");
```
fjtirado: good catch

impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
```diff
@@ -164,9 +186,30 @@ protected CompletableFuture<WorkflowModel> internalExecute(
...
+                      if (cause instanceof TimeoutException) {
+                        return CompletableFuture.failedFuture(
+                            new WorkflowException(
+                                WorkflowError.timeout()
+                                    .instance(taskContext.position().jsonPointer())
+                                    .build(),
+                                cause));
+                      }
+                      return CompletableFuture.failedFuture(e);
+                    });
```
fjtirado: Lets split this logic in two blocks

```suggestion
                .exceptionallyCompose(this::handleTimeoutException);
```

and add 

```
private CompletableFuture<?> handleTimeoutExcepction (Throwable e) {
                      Throwable cause = e instanceof CompletionException ? e.getCause() : e;
                      return CompletableFuture.failedFuture(cause instanceof TimeoutException ? new WorkflowException(WorkflowError.timeout()
                                    .instance(taskContext.position().jsonPointer())
                                    .title(cause.getMessage())
                                    .build(),
                                cause): cause);
}

```

impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
```diff
@@ -164,9 +186,30 @@ protected CompletableFuture<WorkflowModel> internalExecute(
...
       WorkflowContext workflow, TaskContext taskContext, WorkflowModel model) {
     retryIntervalExecutor.ifPresent(r -> r.init(workflow, taskContext, model));
-    return TaskExecutorHelper.processTaskList(
-            taskExecutor, workflow, Optional.of(taskContext), model)
-        .exceptionallyCompose(e -> handleException(e, workflow, taskContext));
+    CompletableFuture<WorkflowModel> future =
+        TaskExecutorHelper.processTaskList(taskExecutor, workflow, Optional.of(taskContext), model);
+    if (attemptDuration.isPresent()) {
+      long timeoutMillis = attemptDuration.get().apply(workflow, taskContext, model).toMillis();
+      if (timeoutMillis > 0) {
```
fjtirado: The double if can be avoided

```suggestion
   long timeoutMillis =  attemptDuration.map(d -> d.apply(workflow,taskContext,model).orElse(Duration.ZERO).toMillis()
    if ( timeoutMillis>0)     {
```

impl/test/src/test/java/io/serverlessworkflow/impl/test/RetryTimeoutTest.java
```diff
@@ -186,6 +187,32 @@ void testRetryEnd() throws IOException {
...
+        app.workflowDefinition(
+                readWorkflowFromClasspath(
+                    "workflows-samples/try-catch-retry-attempt-duration.yaml"))
+            .instance(Map.of())
+            .start();
+    Awaitility.await().atMost(Duration.ofSeconds(5)).until(future::isDone);
+    assertThat(future.join().as(JsonNode.class).orElseThrow()).isEqualTo(result);
+    assertThat(retryListener.taskRetried).hasSize(1);
+    assertThat(retryListener.taskRetried.get("do/0/tryGetPet/try/0/getPet")).isEqualTo((short) 1);
+  }
```
fjtirado, line 233: I do not think is a good idea to catch retry timeout in the test worflow definition, I think we need a test that fails with retry timeout, proving that the task timed out after the specified delay

impl/test/src/test/resources/workflows-samples/try-catch-retry-attempt-duration.yaml
```diff
@@ -0,0 +1,29 @@
...
+            call: http
+            with:
+              method: get
+              endpoint: http://localhost:9797
+              redirect: true
+      catch:
+        errors:
+          with:
+            type: https://serverlessworkflow.io/spec/1.0.0/errors/timeout
+            status: 408
```
fjtirado: I think the exception being catch here should be different than timeout to not mistake with the timeout that should  be launch is the max retry duration is exceeded. 
So use a different one to trigger the retry and to make sure that the overall timeout has worked search for that exception in the unit test

fjtirado (approved):

impl/core/src/main/java/io/serverlessworkflow/impl/executors/TryExecutor.java
```diff
@@ -152,65 +183,123 @@ protected TryExecutor(TryExecutorBuilder builder) {
...
+        duration
+            .map(d -> d.apply(workflowContext, taskContext, model))
+            .orElse(Duration.ZERO)
+            .toMillis();
+    return timeout > 0
+        ? originalFuture
+            .copy()
+            .orTimeout(timeout, TimeUnit.MILLISECONDS)
+            .whenComplete((v, e) -> cancelIfTimeout(e, originalFuture))
+        : originalFuture;
```
fjtirado, line 293: onTimeout do not cancel the underlying future, it just complete it. Since onTimeout do not create a new completable, we need to copy the original one, keep its reference and cancel it.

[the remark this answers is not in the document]
fjtirado: Are you really sure CompletablueFuture.copy() is not available in JDK 17?
Then, how it is possible that the project compiles using JDK 17?
Then, why the Java doc says copy was added in JDK 9? https://docs.oracle.com/javase/9/docs/api/java/util/concurrent/CompletableFuture.html#copy--
If you were a human and report this is a high failure, you will make a lot of good friends.

PR #1605, opened 2026-08-06
[Fix #1604] Shell refactor

### LINKED ISSUE #1599: Incorporate https://github.com/open-workflow-specification/specification/pull/1182 when merged
This pr goal is to restore the Map argument for Script task, 
Currently the schema used for generation is customly modified to keep support for it, once https://github.com/open-workflow-specification/specification/pull/1182 is merged, we can replace the custom schema with the spec schema

### LINKED ISSUE #1604: Drop sh -c in RunShellExecutor, add command allowlist
## Problem

`RunShellExecutor` passes the assembled command string to `ProcessBuilder("sh", "-c", command)`. This means shell metacharacters (`;`, `|`, `&&`, `$(...)`, backticks) in jq-resolved values are interpreted by the shell, making any command allowlist trivially bypassable.

For example, even if only `echo` is allowed:

```yaml
command: ${ "echo " + .user.input }
```

An input of `hello; rm -rf /` becomes `sh -c 'echo hello; rm -rf /'` — the shell executes both commands.

## Proposal

**1. Replace `sh -c` with direct `ProcessBuilder` execution using an argv array.**

The spec already separates `command` (the executable) and `arguments` (the parameters). Map them directly:

- `command` → `argv[0]`
- `arguments` entries → `argv[1..n]` (key-value pairs become `"--key=value"`, key-only becomes `"--key"`)

```yaml
command: echo
arguments:
  --user: john
```

Becomes: `ProcessBuilder("echo", "--user=john")` — no shell involved.

Without a shell, metacharacters are literal strings, not operators. Injection is eliminated structurally, not by filtering.

**2. Add a command allowlist to `WorkflowApplication`.**

```java
WorkflowApplication.builder()
    .allowedShellCommands(List.of("echo", "grep", "curl"))
    .build()
```

- The resolved `command` (argv[0]) is compared against the allowlist as an exact string match.
- Empty list (default) = no shell execution allowed.
- Integrating frameworks pass their configuration through this API.

**3. Pipes, subshells, and redirection are no longer available.**

This is intentional. Users who need pipelines should wrap them in a dedicated script and allowlist that script name. This keeps the trust model clear: deployers control what executables are available, workflow authors control how to invoke them.

## Scope

- `RunShellExecutor` and `RunShellExecutorBuilder` in `impl-core`
- `WorkflowApplication` (new allowlist API)
- `CommandPropertySetter` in `impl-container` is a separate concern (tracked separately)

fjtirado:
Fix https://github.com/open-workflow-specification/sdk-java/issues/1604
Fix https://github.com/open-workflow-specification/sdk-java/issues/1599


types/src/main/resources/schema/workflow.yaml
```diff
@@ -1454,14 +1455,14 @@ $defs:
...
-       format: uri-reference
-       pattern: "^(?!\\s*\\$\\{)(?=\\S)(([^:/?#]+):)?(//([^/?#\\s]*))?([^?#\\s]*)(\\?([^#\\s]*))?(#(\\S*))?$"
+      - title: LiteralUriTemplate
+        type: string
+        format: uri-template
+        pattern: "^(?!\\s*\\$\\{)(?=\\S)(([^:/?#]+):)?(//([^/?#\\s]*))?([^?#\\s]*)(\\?([^#\\s]*))?(#(\\S*))?$"
+      - title: LiteralUri
+        type: string
+        format: uri-reference
+        pattern: "^(?!\\s*\\$\\{)(?=\\S)(([^:/?#]+):)?(//([^/?#\\s]*))?([^?#\\s]*)(\\?([^#\\s]*))?(#(\\S*))?$"
```
fjtirado, line 1465: This appears here as difference because I copied right away the schema from https://github.com/open-workflow-specification/specification/pull/1182. The format (just whitespaces)  of this section was slightly different because of manual modifications I did with previous Jiras (to finally obtain the same schema but with different white spaces). This change is aligning that so both schemas (spec and sdk) will be identical when https://github.com/open-workflow-specification/specification/pull/1182 is merged (which will be hopefully soon)

impl/test/src/test/resources/workflows-samples/run-shell/echo-with-env.yaml
fjtirado, line 1: with this approach we are unavoidily losing the possibility of specifying arguments from environment. I think thats for good, rather than set the enviroment to specify arguments, specify then directly either harcdoded or throoug expressions
The use case that we are not covering is when the user knows that there is an enviroments variable in his SO (a preexisting one, not one we are setting) that should be used as parameter. For covering this case it might be interesting, either manual replacing of the args string after applying the jq expression (replace $PEPE with System.getEnv("PEPE") in the evaluated string) or add the possibility to access the env from jq expressions (which is interesting by itself)
@ricardozanini @mcruzdev In any case, I think is better to adresss that with a different issue

impl/test/src/test/resources/workflows-samples/run-shell/touch-cat.yaml
fjtirado, line 1: I deleted that one, that now has to be implemented as two different consecutive commands. @mcruzdev  its a challenge that you might accept with a different issue, suit yourself ;)

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -49,16 +48,21 @@ public RunShellExecutor(
...
+    if (!workflowContext.definition().application().allowedCommands().contains(command)) {
+      return CompletableFuture.failedFuture(
+          new IllegalAccessException(
+              "Command "
+                  + command
+                  + " is not allowed. Please verify the set of allowed commands passed to the application"));
     }
-    ProcessBuilder builder = new ProcessBuilder("sh", "-c", commandBuilder.toString());
+
+    List<String> commandAndArgs = new java.util.ArrayList<>();
```
ricardozanini:
```suggestion
    List<String> commandAndArgs = new ArrayList<>();
```

impl/test/src/test/resources/workflows-samples/run-shell/echo-with-env.yaml
ricardozanini: Why can't we have `environment`? Not sure if I understood, since `environment` is part of the definition.

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
```diff
@@ -49,16 +48,21 @@ public RunShellExecutor(
...
-      commandBuilder.append(" ").append(entry.getKey().apply(workflowContext, taskContext, model));
-      entry
-          .getValue()
-          .ifPresent(
-              v -> commandBuilder.append("=").append(v.apply(workflowContext, taskContext, model)));
+
+    String command = shellCommand.apply(workflowContext, taskContext, model);
+    if (!workflowContext.definition().application().allowedCommands().contains(command)) {
+      return CompletableFuture.failedFuture(
+          new IllegalAccessException(
```
ricardozanini:
```suggestion
          new SecurityException(
```

> An IllegalAccessException is thrown when an application tries to reflectively create an instance (other than an array), set or get a field, or invoke a method, but the currently executing method does not have access to the definition of the specified class, field, method or constructor.

https://docs.oracle.com/javase/8/docs/api/java/lang/IllegalAccessException.html

Or we do a new `CommandNotAllowedException`.

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -690,4 +704,8 @@ public AllStrategyCorrelationInfoFactory allStrategyCorrelationInfoFactory() {
   public WorkflowLifeCycleCloudEventFactory lifeCycleCloudEventFactory() {
     return lifeCycleCloudEventFactory;
   }
+
+  public Set<String> allowedCommands() {
+    return allowedCommands;
```
ricardozanini, line 709:
```suggestion
    this.allowedCommands = Set.copyOf(builder.allowedCommands);
```

Let's make it immutable so we avoid noise.

impl/test/src/test/resources/workflows-samples/run-shell/echo-none.yaml
```diff
@@ -1,11 +1,13 @@
...
+  name: echo-none
   version: '0.1.0'
 do:
   - runShell:
       run:
         shell:
-          command: 'echo "Serverless Workflow"'
+          command: echo
+          arguments: 
+            - Javierito rulezzzz!!!!
```
ricardozanini:
```suggestion
            - Open Workflow
```

impl/test/src/test/resources/workflows-samples/run-shell/echo-with-env.yaml
fjtirado: We are still populating the environment of the process with the data provided by the definition, but since theere is not longer a shell invocation, if we use enviroment data in the arguments of the invocation, it does not get replaced by the environment value as before. 

Now

```

        command: echo 
        arguments: 
            - Hello $FIRST_NAME $LAST_NAME from env!
        environment:
            FIRST_NAME: John
            LAST_NAME: ${.lastName}
```

Will print in console 
`"Hello $FIRST_NAME $LAST_NAME from env!"`

while deleted test was printing: 

`"Hello John Doe from env!"`

(there was a lastName Doe in the jq input)

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
fjtirado: Yes,  I understood why AI freaks out. that is an illegal attemp to execute a process, so an illegal access exception made perfect sense to me. 
SecurityException is better though (because IllegalAccessException inherits ReflectiveOperationException). I do not think a new exception is needed.

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
fjtirado: I make it inmutable. The suggested change is not just making it inmutable, it is copying it (a thing like the AI love to do, because it is generically "safer", but in our scenario is not needed at all)

impl/test/src/test/resources/workflows-samples/run-shell/echo-none.yaml
fjtirado: Ok, suggested change incorporated ;). I was taking advantage of the fact that the test does not care about the workflow output to make an homage to the author of the review

types/src/main/resources/schema/workflow.yaml
```diff
@@ -924,10 +924,11 @@ $defs:
                       title: ShellStdin
                       description: A runtime expression, if any, to the shell command as standard input (stdin).
                     arguments:
-                      type: object
+                      type: array
                       title: ShellArguments
-                      description: A list of the arguments of the shell command to run.
-                      additionalProperties: true
+                      description: A list of the arguments, if any, of the shell command to run.
+                      items:
```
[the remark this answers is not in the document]
fjtirado: Yes, I hope you realize that before the partial revets the backward compatibility was even worse. 
However after a lot of discussion, it is better to have list rather than map to shell, so the cost of backward incompatibility is assumed

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
```diff
@@ -147,6 +149,7 @@ private WorkflowApplication(Builder builder) {
     this.allStrategyCorrelationInfoFactory = builder.allStrategyCorrelationInfoFactory;
     this.lifeCycleCloudEventFactory = builder.lifeCycleCloudEventFactory;
     this.schedulerExecutorService = builder.schedulerExecutorService;
+    this.allowedCommands = Collections.unmodifiableSet(builder.allowedCommands);
```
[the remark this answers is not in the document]
fjtirado: Im really tired of you complaining when there is not defensive copy. 
this is called at application startup, the wrapped list is a goner not accesible by anyone. 
This being high is plain stupid

impl/test/src/test/resources/workflows-samples/run-shell/echo-jq.yaml
```diff
@@ -1,11 +1,13 @@
...
   namespace: test
-  name: run-shell-example
+  name: echo-jq
   version: '0.1.0'
 do:
   - runShell:
       run:
         shell:
-          command: ${ "echo Hello, \(.user.name)" }
+          command: echo
```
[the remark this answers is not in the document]
fjtirado: how the hell this is high if the test is working?, this is a low, and actually Im not going to change, it is good that if HUMAN user add an extra whitespace, the command still works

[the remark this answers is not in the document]
fjtirado: And if it is working because the command text is trimmed (as proven by the test)

impl/core/src/main/java/io/serverlessworkflow/impl/WorkflowApplication.java
ricardozanini: hmm I don't remember seeing this in the other commit:

>     this.allowedCommands = Collections.unmodifiableSet(builder.allowedCommands);

ricardozanini (approved):

impl/core/src/main/java/io/serverlessworkflow/impl/executors/RunShellExecutor.java
ricardozanini: Hmm it was not AI, I freaked out :D

PR #1650, opened 2026-09-02
Fix oauth2 blocking token fetch

mcruzdev:

This pull request tries to solve the following issue https://github.com/quarkiverse/quarkus-flow/issues/923 in downstream project.


impl/http/src/main/java/io/serverlessworkflow/impl/executors/http/auth/JaxRSAccessTokenProvider.java
```diff
@@ -57,18 +57,21 @@ class JaxRSAccessTokenProvider implements AccessTokenProvider {
...
+            jwt.issuer()
+                .ifPresent(
+                    issuer -> {
+                      if (!issuers.contains(issuer)) {
+                        throw new IllegalStateException("Token issuer is not valid: " + issuer);
+                      }
+                    });
+          }
+          return jwt;
+        });
```
fjtirado: You need to pass the executor. Also this was sync on purpose, the http request is already executed in a different thread. The completable future was intended for reactive REST client APIS. Im surprises QFlow is still calling this code after recent OIDC changes. 
Anyway, provide the executor and I will merge.

fjtirado (approved):

PR #1653, opened 2026-09-02
Update codemodel dependency to 4.0.9.

baldimir:

Updates codemodel dependency to the newest version. The dependency moved to a new groupId. 

I also moved the version definition of dependencies from the jackson module to the parent pom, so it is easier to maintain those versions.


ricardozanini (approved):
LGTM waiting for @mcruzdev report.

fjtirado (approved):

PR #1686, opened 2026-09-17
 [Fix #1685] Adding missing native classes 

### LINKED ISSUE #1685: Missing for native JacksonModelCollectionSerializer and deserializer
**What happened**:

In native mode, serialization fails with errors similar to:

```
InvalidDefinitionException: 
Class io.serverlessworkflow.impl.model.jackson.JacksonModelCollectionSerializer 
has no default (no arg) constructor
```

This appears to be a native image, in which case you may need to configure
reflection for the class that is to be serialized.

Native Image cannot discover unless the corresponding classes are registered for reflection.

Add GraalVM Native Image reflection metadata to the SDK modules that contain classes accessed dynamically by Jackson Collections during serialization and deserialization.

**How to reproduce it**:

Quarkus-Flow Openshift Testsuite agentic-ai tests in native mode can reproduce it.

fjtirado:
Fix https://github.com/open-workflow-specification/sdk-java/issues/1685


gmunozfe (approved):
Looks good to me.
Tested and that exception is gone. Well done @fjtirado

Issue #1695, opened 2026-09-18
Implement optional directory property on the RunShell


cvgaviao:
**What would you like to be added**:
A proposal to improve the specification was made here:
[#1191](https://github.com/open-workflow-specification/specification/issues/1191)

We should implement it in the Java SDK

**Why is this needed**:
Java's java.lang.ProcessBuilder has an explicit method for setting the working directory: directory(File directory).

PR #1696, opened 2026-09-18
Support RunShell's directory property from spec

cvgaviao:
https://github.com/open-workflow-specification/specification/issues/1191

Some shell scripts use the directory where they started (the working directory) to calculate other paths relative to it.
So, we can't always use the directory where the workflow application was started as the script's working directory. If we do, the script will fail.

Java's java.lang.ProcessBuilder has an explicit method to set the working directory: directory(File directory).


fjtirado (commented):
Looks good, but I think we should wait to the result of the discussion on the spec forum. 
In case it is rejected for some reason, since this functionality I agree is needed, this can become a runtime property ( a global one, so all scripts run on the same working directory)

mcruzdev (approved):

matheusandre1:
The pull request is related to #1695; the issue should be linked to the pull request.

